PR #10, opened 2020-10-20
Some cleanups to troi.py entrypoint

alastair:
Some things that I ran into while trying to run patches

### DIFF
--- a/troi.py
+++ b/troi.py
@@ -30,6 +30,13 @@ def playlist(patch, args, debug):
     patch = patches[patch]()
     inputs = patch.inputs()
 
+    required_inputs = [input for input in inputs if not input['optional']]
+    num_required_inputs = len(required_inputs)
+    if num_required_inputs > len(args):
+        input_list = ', '.join([i['name'] for i in required_inputs])
+        print(f"Patch requires {num_required_inputs} inputs ({input_list}) but {len(args)} provided", file=sys.stderr)
+        sys.exit(1)
+
     checked_args = []
     for i, input in enumerate(inputs):
         if not input['optional'] and args[i] is None:
--- a/troi.py
+++ b/troi.py
@@ -83,7 +83,7 @@ def list():
 @click.argument("patch", nargs=1)
 def info(patch):
     """Get info for a given patch"""
-    patches = troi.util.discover_patches("patches")
+    patches = troi.utils.discover_patches("patches")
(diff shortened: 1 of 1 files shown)


troi.py
```diff
@@ -18,30 +17,39 @@ def cli():
...
+        print("Cannot load patch '%s'. Use the list command to get a list of available patches." % patch,
+              file=sys.stderr)
+        sys.exit(1)
 
     patch = patches[patch]()
     inputs = patch.inputs()
 
+    required_inputs = [input for input in inputs if not input['optional']]
+    num_required_inputs = len(required_inputs)
+    if num_required_inputs > len(args):
```
alastair, line 35: I see that you removed this in c6df376ba6, so I assume that there's still a bug that I ran into. I was finding that if I run a patch that has 2 required arguments and I don't specify any, `args` below is empty, and so `args[i]` fails. Will look into it again in more detail before merging this

troi.py
```diff
@@ -18,30 +17,39 @@ def cli():
 @click.argument('args', nargs=-1)
 @click.option('--debug', '-d', is_flag=True, default=False)
 def playlist(patch, args, debug):
-
+    """Generate a playlist using a patch"""
```
alastair, line 20: this adds documentation to the output of `troi.py --help`

troi.py
```diff
@@ -18,30 +17,39 @@ def cli():
...
     if debug:
-        print("- debug mode on")
+        print("- debug mode on", file=sys.stderr)
 
     patches = troi.utils.discover_patches("patches")
-    if not patch in patches:
-        print("Cannot load patch '%s'. Use the list command to get a list of available patches." % patch)
-        quit()
```
alastair, line 28: from my understanding, quit() should only be used in an interactive shell. Since these are error cases I uses sys.exit(1) to fail with an error code

troi.py
```diff
@@ -87,7 +97,7 @@ def info(patch):
     print()
     print("  expected inputs:")
     for input in inputs:
-        print("     %s, type %s" % (input[1], input[0]))
+        print(f"     {input['name']}, type {input['type']}: {input['desc']}")
```
alastair, line 100: I saw that you used python 3.8 in the docker image, so took a risk and used f-strings. Let me know what you think about standardising on these

mayhem (approved):
looks good!

PR #11, opened 2020-10-20
Annoy similarity

alastair:
Initial version that just hits AB, we had some more ideas to add to it:

* filter by similar artists to the input recording
* filter recordings from a given year/range

Should be merged after #10

### DIFF
--- a/troi/acousticbrainz/tests/test_annoy.py
+++ b/troi/acousticbrainz/tests/test_annoy.py
@@ -4,6 +4,7 @@ import unittest.mock
 
 import troi
 import troi.acousticbrainz.annoy
+from troi import PipelineError
 
 return_json = [
   {
@@ -39,7 +40,7 @@ return_json = [
 ]
 
 
-class TestAreaRandomRecording(unittest.TestCase):
+class TestAnnoyLookupElement(unittest.TestCase):
 
     @unittest.mock.patch('requests.get')
     def test_read(self, req):
@@ -61,3 +62,7 @@ class TestAreaRandomRecording(unittest.TestCase):
             "offset": 0
         }
         assert entities[0].mbid == "7b3ecb51-919b-494d-8085-47e3390dd212"
+
+    def test_invalid_metric(self):
+        with self.assertRaises(PipelineError):
+            troi.acousticbrainz.annoy.AnnoyLookupElement("foo", "8f8cc91f-0bca-4351-90d4-ef334ac0a0cf")
--- a/troi/acousticbrainz/tests/test_annoy.py
+++ b/troi/acousticbrainz/tests/test_annoy.py
@@ -1,0 +1,63 @@
+import json
+import unittest
+import unittest.mock
+
+import troi
+import troi.acousticbrainz.annoy
+
+return_json = [
+  {
+    "distance": 0.00010826535435626283,
+    "offset": 0,
+    "recording_mbid": "7b3ecb51-919b-494d-8085-47e3390dd212"
+  },
+  {
+    "distance": 0.3184245228767395,
+    "offset": 0,
+    "recording_mbid": "724335d8-4ae6-4b2d-8be5-056944b8132d"
+  },
+  {
+    "distance": 0.3630277216434479,
+    "offset": 0,
+    "recording_mbid": "d4b46b96-ab56-4f99-a59e-bab590deed8f"
+  },
+  {
+    "distance": 0.3671037256717682,
+    "offset": 0,
+    "recording_mbid": "441a879f-4b8c-4d52-8a6f-e289bef2fd83"
+  },
+  {
+    "distance": 0.3676069676876068,
+    "offset": 0,
+    "recording_mbid": "b26b7d9f-dd39-4b74-89f0-a534d6bbf556"
+  },
+  {
+    "distance": 0.385963499546051,
+    "offset": 0,
+    "recording_mbid": "c1d750f3-93ea-4e0e-85f0-8baa8548c97a"
+  },
+]
+
+
+class TestAreaRandomRecording(unittest.TestCase):
+
+    @unittest.mock.patch('requests.get')
+    def test_read(self, req):
+
+        mock = unittest.mock.MagicMock()
+        mock.status_code = 200
+        mock.text = json.dumps(return_json)
+        req.return_value = mock
+        e = troi.acousticbrainz.annoy.AnnoyLookupElement("mfccs", "8f8cc91f-0bca-4351-90d4-ef334ac0a0cf")
+
+        entities = e.read([[]])
+        req.assert_called_with(e.SERVER_URL + "mfccs/8f8cc91f-0bca-4351-90d4-ef334ac0a0cf",
+                               params={"remove_dups": "true"})
+
+        assert len(entities) == 6
+        assert entities[0].acousticbrainz == {
+            "similarity_from": "8f8cc91f-0bca-4351-90d4-ef334ac0a0cf",
+            "similarity": 0.00010826535435626283,
+            "offset": 0
+        }
+        assert entities[0].mbid == "7b3ecb51-919b-494d-8085-47e3390dd212"
--- a/troi/filters.py
+++ b/troi/filters.py
@@ -115,13 +115,11 @@ class ArtistCreditLimiterElement(troi.Element):
 
 
 class DuplicateRecordingFilterElement(troi.Element):
-    def __init__(self):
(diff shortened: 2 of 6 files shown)


troi/filters.py
```diff
@@ -110,3 +111,32 @@ def read(self, inputs, debug=False):
                 results.append(r)
 
         return results
+
+
+class DuplicateRecordingFilterElement(troi.Element):
```
alastair, line 116: We added a filter to AB so that it would remove items that had the same mbid and same distance score (but a different offset).
What I found is that we have some duplicate submissions with a very similar (but not exact) score. 
This filter was an easy way to remove these dups, and could be useful in other pipelines. We should also add another option to AB to filter these out as well

patches/ab_similar_recordings.py
```diff
@@ -0,0 +1,36 @@
...
+from troi.acousticbrainz.annoy import AnnoyLookupElement
+
+
+class ABSimilarRecordingsPatch(troi.patch.Patch):
+
+    @staticmethod
+    def inputs():
+        return [
+            {"type": str, "name": "recording_id", "desc": "MusicBrainz recording id", "optional": False},
+            {"type": str, "name": "similarity_type", "desc": "Type of acousticbrainz similarity", "optional": False}]
```
mayhem, line 13: It would be really good to list the possible similarities in the message or perhaps give a link to where they can be viewed.

mayhem:
My first attempt at running the code gets me:

`(.ve) mayhem:~/metabrainz/troi-recommendation-playground->./troi.py playlist ab-similar-recordings 145f5c43-0ac2-4886-8b09-63d0e92ded5d mfccs
  annoy: read 151 recordings
Traceback (most recent call last):
  File "./troi.py", line 99, in <module>
    cli()
  File "/Users/robert/metabrainz/troi-recommendation-playground/.ve/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/robert/metabrainz/troi-recommendation-playground/.ve/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/robert/metabrainz/troi-recommendation-playground/.ve/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/robert/metabrainz/troi-recommendation-playground/.ve/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/robert/metabrainz/troi-recommendation-playground/.ve/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "./troi.py", line 53, in playlist
    playlist.generate(debug)
  File "/Users/robert/metabrainz/troi-recommendation-playground/troi/__init__.py", line 48, in generate
    recordings = self.read(source_lists, debug)
  File "/Users/robert/metabrainz/troi-recommendation-playground/troi/playlist.py", line 31, in read
    artist_mbids = [ str(mbid) for mbid in e.artist.mbids or [] ]
AttributeError: 'NoneType' object has no attribute 'mbids'`

mayhem:
Upon first running, I get this error:
https://gist.github.com/mayhem/6cc3018bc295a0fbeb4e49dc0daabb4b

alastair:
Yeah, I need to open an issue for that. There seems to be a bug in the recording lookup element that leaves a recording in the pipeline if it can't find metadata, and then the playlist element tries to read non-existent metadata

patches/ab_similar_recordings.py
alastair: I was thinking about this. There's also a check here: https://github.com/metabrainz/troi-recommendation-playground/pull/11/files#diff-27abf483d83da8da4f40516af583a0e57706a1c0b345c06b22b54b2f5783073dR27, so it'd be great to be able to show it in the help without having to duplicate these constants. I'll see if argparse can help us here

alastair:
This has most of the items that we discussed over the last few weeks:

 * Allow all types of similarity
 * Add tests to annoy element
 * Add example patch that includes this element
 
For now, leaving out filtering by year (waiting on fixes to the lookup).

troi/acousticbrainz/tests/test_annoy.py
```diff
@@ -0,0 +1,63 @@
...
+  },
+  {
+    "distance": 0.385963499546051,
+    "offset": 0,
+    "recording_mbid": "c1d750f3-93ea-4e0e-85f0-8baa8548c97a"
+  },
+]
+
+
+class TestAreaRandomRecording(unittest.TestCase):
```
mayhem, line 42: Time to update the class name?

troi/filters.py
```diff
@@ -111,3 +112,56 @@ def read(self, inputs, debug=False):
...
+    """This Element takes a list of recordings and removes any duplicate recordings
+    that appear directly after each other (based on recording MBID)
+    If recordings are in the order mbid1, mbid2, mbid1, this will not be considered
+    a duplicate.
+    """
```
mayhem, line 122: This is a very oddly specific filter -- I think the name of it is currently misleading. Judging by the name I would expect it to filter out all duplicates. Maybe call it ConsecutiveRecordingFilterElement()? Or maybe add the functionality to remove all duplicates while preserving the recording order?

troi/patches/ab_similar_recordings.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @staticmethod
+    def description():
+        return "Find acoustically similar recordings from AcousticBrainz"
+
+    def create(self, inputs):
+        recording_id = inputs[0]
+        similarity_type = inputs[1]
+
+        if similarity_type not in VALID_METRICS:
```
mayhem, line 31: Maybe make this check case insensitive?

troi/tests/test_utils.py
```diff
@@ -9,7 +9,7 @@ class TestPatches(unittest.TestCase):
     def test_discover_patches(self):
         patches = discover_patches()
 
-        assert len(patches) == 2
+        assert len(patches) == 3
         assert "daily-jams" in patches
         assert "area-random-recordings" in patches
```
mayhem, line 15: Add an assert that ab-similar-recordings is in patches

mayhem (changes requested):
Almost good to go, minor tweaks.

troi/acousticbrainz/tests/test_annoy.py
alastair: doh

troi/filters.py
alastair: Agreed, being able to to filter all items preserving order is a good idea. I guess this is the same behaviour as `uniq`!

troi/patches/ab_similar_recordings.py
alastair: sure. do you think this check makes more sense here or in `AnnoyLookupElement`? (or maybe in both?)

mayhem: Both seems overkill, but in the Element makes sense to me.

troi/filters.py
mayhem: I think in general Elements/Filters should preserve order -- while we don't rely on that now, I think it is a good goal to shoot for -- I suspect it will be useful later.

alastair: after discussion, we added 2 filters for each of these behaviours.
Was also missing tests, so added those.

troi/patches/ab_similar_recordings.py
alastair: added this, and moved check to the element

alastair:
ready!

troi/patches/ab_similar_recordings.py
```diff
@@ -0,0 +1,45 @@
+import troi.patch
+import troi.filters
+import troi.listenbrainz.recs
+import troi.musicbrainz.recording_lookup
+from troi import PipelineError
+from troi.acousticbrainz.annoy import AnnoyLookupElement
+
+
+VALID_METRICS = ['mfccs', 'mfccsw', 'gfccs', 'gfccsw', 'key', 'bpm', 'onsetrate', 'moods',
```
mayhem, line 9: why are these duplicated here? why not just import from the element?

mayhem (changes requested):
Minor nitpicks, mostly, but quite close!

mayhem (approved):

alastair:
crap, forgot the failing test. fixing now

PR #12, opened 2020-10-20
Initial version of docs

alastair:
LB-737
This is just a stub docs directory so that we can build it and publish on readthedocs. We need to continue by adding an intro and linking into the existing files.


mayhem (approved):
:shipit:

PR #13, opened 2020-10-20
Add a more detailed setup.py for upload to pypi

alastair:
LB-739
This is an initial version of setup.py that allows us to upload troi to pypi. We should still address some of the points raised in the above ticket in the future.


mayhem (approved):
:cake:

PR #14, opened 2020-10-21
Exception cleanup (fixing LB-743)

mayhem:
As per our discussion in IRC.

### DIFF
--- a/troi/acousticbrainz/annoy.py
+++ b/troi/acousticbrainz/annoy.py
@@ -5,7 +5,7 @@ from urllib.parse import quote
 import requests
 import ujson
 
-from troi import Element, Recording
+from troi import Element, Recording, PipelineError
 
 
 class AnnoyLookupElement(Element):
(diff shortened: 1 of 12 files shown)


troi/acousticbrainz/annoy.py
```diff
@@ -34,12 +34,12 @@ def read(self, inputs, debug=False):
         url = self.SERVER_URL + self.metric + "/" + self.mbid
         r = requests.get(url)
         if r.status_code != 200:
-            r.raise_for_status()
+            raise PipelineError("Cannot fetch annoy similarities from AcousticBrainz: HTTP code %d" % r.status_code)
```
alastair, line 37: I think this is an OK way of doing it. raise_for_status just checks the status code and raises an exception if it's not 200. This means the if/raise here is redundant.
It should be no different here to do if/raise PiplineError or try/raise_for_status/except/raise PipelineError, and the first one is less lines, so let's stay with that

troi/musicbrainz/artist_credit_id_lookup.py
```diff
@@ -38,12 +38,12 @@ def read(self, inputs, debug=False):
 
         r = requests.get(url)
         if r.status_code != 200:
-            r.raise_for_status()
+            raise PipelineError("Cannot fetch artist credits from ListenBrainz: HTTP code %d" % r.status_code)
```
alastair, line 41: musicbrainz? Or is this ListenBrainz because of the API that it's hitting?

troi/playlist.py
```diff
@@ -32,7 +32,7 @@ def read(self, inputs, debug=False):
             listens.append({
```
alastair, line 32: there's a potential error on the line above this (I can't annotate it in this pr), where you call `e.artist.mbids`, but `e.artist` could be None

alastair (approved):
Looks fine. We'll probably find that there will be some additional exceptions that we could catch that might fall out in the future, but OK for now.

troi/musicbrainz/artist_credit_id_lookup.py
mayhem: The latter -- its hitting an LB api, this error message LB.

PR #15, opened 2020-10-21
Setup improvements (LB-739)

mayhem:
Improve setup of the python module as described in the ticket and discussed in IRC.


alastair (approved):
perfect 👍

PR #16, opened 2020-10-21
Add support for loading patches from the patches directory in the current directory.

mayhem:
As it says on the tin.

### DIFF
--- a/troi/utils.py
+++ b/troi/utils.py
@@ -9,12 +9,20 @@ import troi.patch
 
 
 def discover_patches():
+    """
+        Attempt to load patches from the installed patches dir as well as any patches directory in the current dir.
+    """
+
     patches = discover_patches_from_dir("troi.patches.", os.path.join(os.path.dirname(__file__), "patches"))
     local_patches = discover_patches_from_dir("patches.", "./patches", True)
     return  {**patches, **local_patches}
 
 
 def discover_patches_from_dir(module_path, patch_dir, add_dot=False):
+    """
+        Load patches given the appropriate python module path and then file system path. 
+        If add_dot = True, add . to the sys.path and then remove it before this function exists.
+    """
 
     if add_dot:
         sys.path.append(".")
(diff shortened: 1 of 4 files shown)


troi/utils.py
```diff
@@ -2,12 +2,30 @@
...
 import troi.patch
 
 
-def discover_patches(patch_dir):
+def discover_patches():
+    """
+        Attempt to load patches from the installed patches dir as well as any patches directory in the current dir.
+    """
+
+    patches = discover_patches_from_dir("troi.patches.", os.path.join(os.path.dirname(__file__), "patches"))
```
alastair, line 16: it's a bit annoying that you have to specify this twice. Any thoughts on how you could specify it just once?
For example you could do:

```
>>> patches = importlib.import_module("troi.patches")
>>> patches.__file__
'/Users/alastair/code/metabrainz/troi-recommendation-playground/troi/patches/__init__.py'
```

which you could then use os.path.dirname in order to get the dir to use `listdir`.
This requires that it's a module (i.e. has `__init__.py`)

troi/utils.py
```diff
@@ -2,12 +2,30 @@
...
+    """
+        Load patches given the appropriate python module path and then file system path. 
+        If add_dot = True, add . to the sys.path and then remove it before this function exists.
+    """
+
+    if add_dot:
+        sys.path.append(".")
 
     patch_dict = {}
     for path in os.listdir(patch_dir):
```
alastair, line 31: using `glob.glob("*.py")` means that you wouldn't have to do specific checks for the extension later here.
I think that os.listdir in python [doesn't return . and ..](https://docs.python.org/3/library/os.html#os.listdir), this isn't C 😄

alastair (approved):
Looks good, a few small cleanups.

troi/utils.py
mayhem: I'm skipping the change to glob -- it makes things more complicated. I'll remove the C like checks, sure...

troi/utils.py
mayhem: Mildly annoying yes. Not worth sticking more time into it -- I've already spent loads of time on this code and don't really feel like improving tiny bits like this.

PR #19, opened 2020-10-23
Adding more tests

mayhem:
This adds tests for core functionality. At this point the most important modules/features seem covered -- we should discuss what other modules/functions need more tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ setup(
         "License :: OSI Approved :: MIT License",
         "Operating System :: OS Independent",
     ],
-    python_requires='>=3.7',
+    python_requires='>=3.6',
     entry_points = {
         'console_scripts': ['troi=troi.cli:cli'],
     }
--- a/troi/musicbrainz/tests/test_related_artist_credits.py
+++ b/troi/musicbrainz/tests/test_related_artist_credits.py
@@ -1,68 +1,55 @@
-import json
 import os
 import unittest
 import unittest.mock
 
 import troi
-import troi.musicbrainz.recording_lookup
+import troi.musicbrainz.related_artist_credits
 
-request_json = [
-                    { "[recording_mbid]": "a96bf3b6-651d-49f4-9a89-eee27cecc18e" }, 
-                    { "[recording_mbid]": "ec5b8aa9-7483-4791-a185-1f599a0cdc35" }
-               ]
-
-return_json = [
+return_json = """[
     {
-        "[artist_credit_mbids]": [
-            "8f6bd1e4-fbe1-4f50-aa9b-94c450ec0f11"
-        ],
         "artist_credit_id": 65,
         "artist_credit_name": "Portishead",
-        "comment": "",
-        "length": 253000,
-        "recording_mbid": "a96bf3b6-651d-49f4-9a89-eee27cecc18e",
-        "recording_name": "Sour Times",
-        "original_recording_mbid": "a96bf3b6-651d-49f4-9a89-eee27cecc18e"
+        "count": 64,
+        "related_artist_credit_id": 20660,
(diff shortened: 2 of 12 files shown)


setup.py
```diff
@@ -18,7 +18,7 @@
         "License :: OSI Approved :: MIT License",
         "Operating System :: OS Independent",
     ],
-    python_requires='>=3.7',
+    python_requires='>=3.6',
```
alastair, line 21: This has gone down from 3.8 to 3.7 to 3.6. I'm just curious to know why - is there a reason that you need an older version, or are you just interested in supporting all currently supported pythons?

troi/musicbrainz/tests/test_related_artist_credits.py
```diff
@@ -1,68 +1,55 @@
...
-import troi.musicbrainz.recording_lookup
+import troi.musicbrainz.related_artist_credits
 
-request_json = [
-                    { "[recording_mbid]": "a96bf3b6-651d-49f4-9a89-eee27cecc18e" }, 
-                    { "[recording_mbid]": "ec5b8aa9-7483-4791-a185-1f599a0cdc35" }
-               ]
-
-return_json = [
+return_json = """[
```
alastair, line 8: this is a string but others in `test_recording_lookup.py` and `test_area_random_recordings.py` use an object and json.dumps when setting the mock's return value

troi/tests/test_utils.py
```diff
@@ -0,0 +1,18 @@
...
+class TestPatches(unittest.TestCase):
+
+    def test_discover_patches(self):
+        patches = discover_patches()
+
+        assert len(patches) == 2
+        assert "daily-jams" in patches
+        assert "area-random-recordings" in patches
+
+        print(patches)
```
alastair, line 16: extra print

troi/tests/test_utils.py
```diff
@@ -0,0 +1,18 @@
...
+from troi.utils import discover_patches
+from troi.patch import Patch
+
+
+class TestPatches(unittest.TestCase):
+
+    def test_discover_patches(self):
+        patches = discover_patches()
+
+        assert len(patches) == 2
```
alastair, line 12: are you expecting to have to change this every time we add a new built-in patch?

alastair (approved):
Tests look fine here 👍

setup.py
mayhem: My mac still has 3.6...

troi/tests/test_utils.py
mayhem: Yes. My intention was to be a bit more explicit about adding new patches, but I don't feel strongly about it.

setup.py
alastair: given that we're not focusing on building in docker, that's a perfectly fine reason.

troi/tests/test_utils.py
mayhem: Fixed and committed to main, I didn't see these comments at first.

troi/musicbrainz/tests/test_related_artist_credits.py
mayhem: Ah, yes. Fixed.

troi/tests/test_utils.py
alastair: not sure if you prefer that I mark reviews as changes requested even when there are only small changes to be made? I didn't think that it was worth doing 2 rounds for such a small change. Or I can be more explicit on the review message.

mayhem: I think if you used "request changes" if there are issues that you really want to see fixed and "approve" if you're fine with merging, regardless of what comments you've left, I think that would work well for me. WDYT?

PR #20, opened 2020-10-26
Pass all test subcommand arguments and options to pytest

alastair:
Took a while to realise that the reason that `list(x)` wasn't working is that there was a function in this file called `list`. This command will now be `troi list-patches`. We can rename the method using click if you want to keep it `list`.


alastair:
I force pushed to make the command `list` again

mayhem (approved):
LGTM

PR #21, opened 2020-10-27
LB-740: Implement logging functions

mayhem:
WIP -- logging statements added as an example. Current output:

INFO:DailyJamsPatch:this is a log message
INFO:DailyJamsElement:element log message
 . .

### DIFF
--- a/troi/patch.py
+++ b/troi/patch.py
@@ -1,3 +1,4 @@
+import logging
 from abc import ABC, abstractmethod
 
 import troi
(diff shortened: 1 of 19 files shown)


troi/patch.py
```diff
@@ -1,12 +1,33 @@
...
 
-    def __init__(self):
-        pass
+    def __init__(self, debug=False):
+        if debug:
+            level = logging.DEBUG
+        else:
+            level = logging.INFO
+        logging.basicConfig(level=level)
+        self.logger = logging.getLogger(type(self).__name__)
```
alastair, line 15: I was looking at the logging docs, loggers are hierarchical with . as a separator. Therefore there might be value in using `type(self)` as the logger. This would allow us to configure a logger `troi`, and have it apply to all child loggers. Theoretically this would allow us to do things like having different logging levels for elements and patches, although I'm not sure if this is something that we have a specific requirement for

alastair (commented):
Some python 3 improvements that we can make.

Also, I think that we can simplify the configuration of debug mode, without having to pass a flag into the classes - see comments below

troi/patch.py
```diff
@@ -1,12 +1,33 @@
...
 from abc import ABC, abstractmethod
 
 import troi
 
 
 class Patch(ABC):
 
-    def __init__(self):
-        pass
+    def __init__(self, debug=False):
```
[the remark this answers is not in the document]
mayhem: I forgot to remove this leftover -- no longer needed since we have the debug statement. Should we need something fancier we can add it later.

troi/patch.py
mayhem: I don't see it for right now -- lets keep it simple and if we learn we need better debug support, we'll add it.

alastair (approved):
yes! much cleaner without `debug` everywhere

alastair:
@brainzbot test this please

PR #23, opened 2020-10-28
Use request's params argument instead of encoding manually

alastair:
Quick fix to use `params` to pass in a dict of query parameters instead of constructing it manually:

* Requests quotes things for you, so `quote()` isn't needed
* Easier to change parameters if needed without having to faff about with ? and &
* use `r.url` after the request has been made to see the full url with query params if needed


mayhem (approved):
much nicer!

PR #27, opened 2020-11-02
Code cleanup / linting

alastair:
Some warnings and errors that my code checker was throwing up in the code

* A bunch of unused imports
* Some missing imports in code flows that we never hit
* Remove debug parameter from Element.read subclasses
* Clean up some arguments in subclasses, especially in @staticmethod-annotated methods
* Remove some unused code
* Should use `is` to check None, not equal
* Use `None` as a default empty argument, [instead of mutable `{}`](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)

### DIFF
--- a/troi/__init__.py
+++ b/troi/__init__.py
@@ -11,20 +11,17 @@ class Element(ABC):
         self.sources = []
         self.logger = logging.getLogger(type(self).__name__)
 
-
     def log(self, msg):
         '''
             Log a message with the info log level, which is the default for troi.
         '''
         self.logger.info(msg)
 
-
     def debug(self, msg):
         '''
             Log a message with debug log level. These messages will only be shown when debugging is enabled.
         '''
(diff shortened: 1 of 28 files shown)


troi/__init__.py
```diff
@@ -97,7 +94,7 @@ def outputs():
         return None
 
     @abstractmethod
-    def read(self, source_data_list, debug):
+    def read(self, source_data_list):
```
mayhem, line 97: I still missed some? Sheesh!

mayhem (approved):

troi/__init__.py
alastair: I think I got them all this time. It still exists in the constructor of Patch, which I believe is the only place that it needs to be, right?

mayhem: Correct.

PR #28, opened 2020-11-10
Remove runtimeerror

alastair:
Remove RuntimeError exception in favour of our troi-specific subclass, PipelineError

Small fix for a failing annoy test that I merged in #11


mayhem (approved):
Thanks!

PR #29, opened 2020-11-11
Add a year lookup element

mayhem:
Not much to say, there should be no real surprises here.

### DIFF
--- a/troi/filters.py
+++ b/troi/filters.py
@@ -197,12 +197,13 @@ class YearRangeFilterElement(troi.Element):
         Remove/keep recordings if they do or do not belong in a range of years
     '''
 
-    def __init__(self, start_year, end_year, inverse=False):
+    def __init__(self, start_year, end_year=None, inverse=False):
         '''
             Filter a list of Recordings based on their year -- the year must be between
-            start_year and end_year, otherwise the recording will be filtered out.
-            If inverse=True, then keep all Recordings that do no fit into the given year
-            range.
+            start_year and end_year, otherwise the recording will be filtered out. If no
+            end_year is given, keep (or reject in case of inverse) tracks greater or equal
+            to start_year. If inverse=True, then keep all Recordings that do no fit into the
+            given year range.
         '''
         troi.Element.__init__(self)
         self.start_year = start_year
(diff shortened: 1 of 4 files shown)


troi/musicbrainz/year_lookup.py
```diff
@@ -0,0 +1,63 @@
...
+            raise PipelineError("Cannot fetch recording years from MusicBrainz: HTTP code %d" % r.status_code)
+
+        try:
+            rows = ujson.loads(r.text)
+        except ValueError as err:
+            raise PipelineError("Cannot fetch recording years from MusicBrainz: " + str(err))
+
+        mbid_index = {}
+        for row in rows:
+            mbid_index[row['artist_credit_name'] + row['recording_name']] = row['year']
```
alastair, line 51: If you pass in more than 1 AC + recording, and they all fold into a single one, the AC and recording name in the result will only include the last instance of those values. 

e.g. 
AC: Queen,queen; recording: Death on two legs,death-on-two-legs
only returns keys queen, death-on-two-legs

alastair (commented):
Except for the almost-but-not-quite-identical-arguments issue, looks fine.

troi/musicbrainz/year_lookup.py
mayhem: Fixed in bono-data-sets.

troi/filters.py
```diff
@@ -190,3 +190,47 @@ def read(self, inputs, debug=False):
...
 
         return output
+
+
+class YearRangeFilterElement(troi.Element):
+    '''
+        Remove/keep recordings if they do or do not belong in a range of years
+    '''
+
+    def __init__(self, start_year, end_year, inverse=False):
```
alastair, line 200: nice-to-have would be to make end_year optional so that you can say "Give me stuff from 1997" with a single argument, but not important.

troi/filters.py
```diff
@@ -190,3 +190,47 @@ def read(self, inputs, debug=False):
...
+
+    @staticmethod
+    def inputs():
+        return [Recording]
+
+    @staticmethod
+    def outputs():
+        return [Recording]
+
+    def read(self, inputs, debug=False):
```
alastair, line 220: debug isn't needed?

alastair (approved):
Fine here. 2 small comments.

troi/filters.py
mayhem: Fixed.

troi/filters.py
mayhem: Now implemented.

PR #30, opened 2020-11-12
Add YearSortElement

mayhem:
AISOTT.

### DIFF
--- a/troi/sorts.py
+++ b/troi/sorts.py
@@ -1,3 +1,4 @@
+from sys import maxsize
 from operator import attrgetter
 
 import troi
@@ -11,7 +12,11 @@ class YearSortElement(troi.Element):
 
     def __init__(self, reverse=False):
         '''
-            Sort recordings by year ascending. If reverse=True, sort descending.
+            Sort recordings by year ascending -- recordings that have the same
+            year will be returned in an undefined order. Recordings
+            that have no year set will be returned at the end of the list.
+            If reverse=True, sort descending and return tracks with no year
+            first.
         '''
         super().__init__()
         self.reverse = reverse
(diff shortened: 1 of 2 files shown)


troi/sorts.py
```diff
@@ -0,0 +1,34 @@
+from operator import attrgetter
+
+import troi
+from troi import PipelineError, Recording
+
+
+class YearSortElement(troi.Element):
+    '''
+        Sort recordings by year.
```
alastair, line 9: add documentation that says behaviour is undefined for multiple items with the same year

troi/sorts.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @staticmethod
+    def outputs():
+        return [Recording]
+
+    def read(self, inputs):
+
+        def year_sorter(r):
+            if not r.year:
+                return 0
```
alastair, line 31: This sorts items with no year first. Is this desired behaviour? My gut feeling is that they should be at the end.

troi/tests/test_sorts.py
```diff
@@ -0,0 +1,26 @@
+import unittest
+
+from troi import Artist, Recording
+import troi.sorts 
+
+
+class TestYearSortElement(unittest.TestCase):
+
+    def test_year_sort_element(self):
```
alastair, line 9: would be good to see a test with a None year.

alastair (changes requested):

troi/sorts.py
mayhem: I didn't have a preference, so I changed it as you suggested.

troi/sorts.py
mayhem: Done.

alastair (approved):
looks good, thanks.

Eventually we might want to consider a sorting element that allows you to pass in your own sorting key, but that's not important yet.

PR #31, opened 2020-11-19
LB-736: Use click to parse the arguments of patches

alastair:
Instead of setting arguments for a patch by defining a dictionary, use a click command. This has a number of advantages:
 * Use click's robust argument parsing, type checking, and standardised definition structure
 * Automatically generate help, and use it in the `info` command
 * explicitly name inputs in the `create` methods

### DIFF
--- a/troi/patches/ab_similar_recordings.py
+++ b/troi/patches/ab_similar_recordings.py
@@ -1,3 +1,5 @@
+import click
+
 import troi
 import troi.patch
 import troi.filters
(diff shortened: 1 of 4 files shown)


troi/patches/ab_similar_recordings.py
```diff
@@ -6,13 +8,26 @@
...
-            {"type": str, "name": "recording_id", "desc": "MusicBrainz recording id", "optional": False},
-            {"type": str, "name": "similarity_type", "desc": "Type of acousticbrainz similarity", "optional": False}]
+    @cli.command(no_args_is_help=True)
+    @click.argument('recording_id')
+    @click.argument('similarity_type')
+    def parse_args(**kwargs):
+        """
+        Find acoustically similar recordings from AcousticBrainz.
+
+        \b
```
mayhem, line 26: what magic is this?

mayhem (approved):
This is really nice! Looks a bit odd having click stuff in patches, but this clearly simplifies things.

troi/patches/ab_similar_recordings.py
alastair: click magically reflows paragraphs based on the width of your terminal. This forces click to treat the following as verbatim lines

PR #32, opened 2020-11-19
Add element to look up tags and genres on dataset hoster

alastair:
I used the `Recording.musicbrainz` dictionary to store this data, because that's where it comes from.


mayhem (approved):
:shipit:

PR #33, opened 2020-12-04
Oh goodie, another large, rambling PR from rob! 🙄

mayhem:
Way overdue and another long ramble. Sorry!

It brings:
- Playlist entity, 
- JSPF serializing
- LB submission
- weekly flashback jams
- improved command line args for playlist command (token, save, print)
-pylint cleanup

### DIFF
--- a/troi/__init__.py
+++ b/troi/__init__.py
@@ -223,14 +223,13 @@ class Playlist(Entity):
     """
         The class that represents a playlist, which is nothing more than a [Recording] with metadata.
     """
-    def __init__(self, name=None, mbid=None, desc=None, filename=None,
+    def __init__(self, name=None, mbid=None, filename=None, recordings=[],
                  ranking=None, year=None, musicbrainz=None, listenbrainz=None, acousticbrainz=None):
         Entity.__init__(self, ranking=ranking, musicbrainz=musicbrainz, listenbrainz=listenbrainz, acousticbrainz=acousticbrainz)
         self.name = name
-        self.desc = desc
         self.filename = filename
         self.mbid = mbid
-        self.recordings = []
+        self.recordings = recordings
 
     def __str__(self):
         return "<Recording('%s', %s, %s)>" % (self.name, self.desc, self.mbid)
--- a/troi/__init__.py
+++ b/troi/__init__.py
@@ -222,8 +222,12 @@ class Recording(Entity):
 class Playlist(Entity):
     """
         The class that represents a playlist, which is nothing more than a [Recording] with metadata.
+
+        The arguments are the same as the other entities, except that mbid in this case refers to a playlist_mbid
+        and that filename is the suggested filename that this playlist should be saved as, if the user asked to 
+        do that and didn't provide a different filename.
     """
-    def __init__(self, name=None, mbid=None, filename=None, recordings=[],
+    def __init__(self, name=None, mbid=None, filename=None, recordings=None,
                  ranking=None, year=None, musicbrainz=None, listenbrainz=None, acousticbrainz=None):
         Entity.__init__(self, ranking=ranking, musicbrainz=musicbrainz, listenbrainz=listenbrainz, acousticbrainz=acousticbrainz)
         self.name = name
@@ -232,7 +236,7 @@ class Playlist(Entity):
         self.recordings = recordings
 
     def __str__(self):
-        return "<Recording('%s', %s, %s)>" % (self.name, self.desc, self.mbid)
+        return "<Playlist('%s', %s, %s)>" % (self.name, self.desc, self.mbid)
 
 
 
--- a/troi/__init__.py
+++ b/troi/__init__.py
@@ -59,10 +59,10 @@ class Element(ABC):
             for source in self.sources:
                 source_lists.append(source.generate())
 
-        recordings = self.read(source_lists)
-        self.debug("%-50s %d items" % (type(self).__name__[:49], len(recordings or [])))
+        items = self.read(source_lists)
+        self.debug("%-50s %d items" % (type(self).__name__[:49], len(items or [])))
 
-        return recordings
+        return items
 
     def run(self):
         """
--- a/troi/__init__.py
+++ b/troi/__init__.py
@@ -221,7 +221,9 @@ class Recording(Entity):
 
 class Playlist(Entity):
     """
-        The class that represents a playlist, which is nothing more than a [Recording] with metadata.
+        The class that represents a playlist, which is nothing more than a [Recording] with metadata. This is
+        different from the PlaylistElement, which is normally uses to terminate a pipeline for submitting or
+        saving result playlists to disk.
 
         The arguments are the same as the other entities, except that mbid in this case refers to a playlist_mbid
         and that filename is the suggested filename that this playlist should be saved as, if the user asked to 
--- a/troi/__init__.py
+++ b/troi/__init__.py
@@ -219,6 +219,23 @@ class Recording(Entity):
         return "<Recording('%s', %s, %s)>" % (self.name, self.mbid, self.msid)
 
 
+class Playlist(Entity):
+    """
+        The class that represents a playlist, which is nothing more than a [Recording] with metadata.
+    """
+    def __init__(self, name=None, mbid=None, desc=None, 
+                 ranking=None, year=None, musicbrainz=None, listenbrainz=None, acousticbrainz=None):
+        Entity.__init__(self, ranking=ranking, musicbrainz=musicbrainz, listenbrainz=listenbrainz, acousticbrainz=acousticbrainz)
+        self.name = name
+        self.desc = desc
+        self.mbid = mbid
+        self.recordings = []
+
+    def __str__(self):
+        return "<Recording('%s', %s, %s)>" % (self.name, self.desc, self.mbid)
+
+
+
 class PipelineError(RuntimeError):
     """
         An exception to be thrown when the pipeline encounters an erorr that is a runtime error and not a
--- a/troi/__init__.py
+++ b/troi/__init__.py
@@ -223,11 +223,12 @@ class Playlist(Entity):
     """
         The class that represents a playlist, which is nothing more than a [Recording] with metadata.
     """
-    def __init__(self, name=None, mbid=None, desc=None, 
+    def __init__(self, name=None, mbid=None, desc=None, filename=None,
                  ranking=None, year=None, musicbrainz=None, listenbrainz=None, acousticbrainz=None):
         Entity.__init__(self, ranking=ranking, musicbrainz=musicbrainz, listenbrainz=listenbrainz, acousticbrainz=acousticbrainz)
         self.name = name
         self.desc = desc
+        self.filename = filename
         self.mbid = mbid
         self.recordings = []
 
--- a/troi/cli.py
+++ b/troi/cli.py
@@ -64,8 +64,7 @@ def playlist(patch, args, debug):
         return
 
     playlist.print()
-    print("-- generated playlist with %d recordings. Open playlist by opening playlist.html in your browser." % len(playlist.entities))
-    playlist.launch()
+    playlist.save()
 
 
 @cli.command(name="list")
--- a/troi/cli.py
+++ b/troi/cli.py
@@ -23,8 +23,9 @@ def cli():
 @click.option('--print', '-p', 'echo', required=False, is_flag=True)
 @click.option('--save', '-s', required=False, is_flag=True)
 @click.option('--token', '-t', required=False, type=click.UUID)
+@click.option('--created-for', '-c', required=False)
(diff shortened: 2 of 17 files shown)


mayhem:
But, hey look! https://test.listenbrainz.org/playlist/af276fe6-3a64-4e8e-84e8-f3a8adc153cc

troi/__init__.py
```diff
@@ -219,6 +219,23 @@ def __str__(self):
         return "<Recording('%s', %s, %s)>" % (self.name, self.mbid, self.msid)
 
 
+class Playlist(Entity):
+    """
+        The class that represents a playlist, which is nothing more than a [Recording] with metadata.
+    """
+    def __init__(self, name=None, mbid=None, filename=None, recordings=[],
```
alastair, line 226: bad idea to set a default parameter as `[]`, use None instead and in the code do `self.recordings = recordings or []`

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

troi/cli.py
```diff
@@ -20,8 +20,12 @@ def cli():
 ))
 @click.argument('patch', type=str)
 @click.option('--debug/--no-debug')
+@click.option('--print', '-p', 'echo', required=False, is_flag=True)
+@click.option('--save', '-s', required=False, is_flag=True)
+@click.option('--token', '-t', required=False, type=click.UUID)
```
alastair, line 25: I would like to see documentation for these flags. See how I did it for the other patches in my previous PR - click doesn't let add documentation directly to options, so you have to do it by writing a docstring.

troi/cli.py
```diff
@@ -46,9 +50,29 @@ def playlist(patch, debug, args):
...
+    if save:
+        playlist.save()
+        print("playlist saved.")
+
+    if echo:
+        playlist.print()
+
+    if not echo and not save and not token:
+        if len(playlist.playlists) == 0:
+            print("No playlists where generated. :(")
```
alastair, line 69: were

troi/playlist.py
```diff
@@ -1,79 +1,148 @@
...
+            }
+        tracks.append(track)
+
+    data['track'] = tracks
+
+    return { "playlist" : data }
+
+class PlaylistElement(Element):
+    """
+        Take a list of Recordings or Playlists and save, print or submit them.
```
alastair, line 42: would be good to have some documentation that says that this actually stores a _list_ of playlists, and documentation in methods like `save` which say that if there are more than one that they're enumerated.

Add documentation saying that a list of Recordings is treated as a single playlist?

troi/cli.py
```diff
@@ -46,9 +50,29 @@ def playlist(patch, debug, args):
...
+        elif len(playlist.playlists) == 1:
+            print("A playlist with %d tracks was generated." % len(playlist.playlists[0].recordings))
+        else:
+            print("%d playlists were generated." % len(playlist.playlists))
+
+        print("\nBut, you didn't tell me what to do with it, so I discarded it. (hint: use --submit or --print)")
```
alastair, line 75: these flags need to be updated

troi/cli.py
```diff
@@ -46,9 +50,29 @@ def playlist(patch, debug, args):
...
 
-    playlist.print()
-    print("-- generated playlist with %d recordings. Open playlist by opening playlist.html in your browser." % len(playlist.entities))
-    playlist.launch()
+    if token:
+        mbids = playlist.submit(token, created_for)
+        if len(mbids) == 1:
+            print("Submitted playlist: %s" % mbids[0])
+        else:
+            print("Submitted playlists:", mbid)
```
alastair, line 58: should this be `mbids`?
Is this a playlist id? might be nice to prefix it with http://host/playlist for easy clicking

troi/patches/daily_jams.py
```diff
@@ -19,20 +19,21 @@ class DailyJamsElement(Element):
         Split weekly recommended recordings into 7 sets, one for each day of the week.
     '''
 
-    def __init__(self, recs, day):
+    def __init__(self, recs, user, day):
```
alastair, line 22: we haven't been focusing on docstrings for `__init__` in these elements, but I think it's becoming more important and we should start adding brief documentation here

troi/patches/daily_jams.py
```diff
@@ -19,20 +19,21 @@ class DailyJamsElement(Element):
...
         Element.__init__(self)
         self.recs = recs
         self.day = day
+        self.user = user
```
alastair, line 26: small thing, can we order the assignments of these in the same order as they are in the function definition?

troi/patches/daily_jams.py
```diff
@@ -19,20 +19,21 @@ class DailyJamsElement(Element):
         Split weekly recommended recordings into 7 sets, one for each day of the week.
```
alastair, line 19: what do we think about putting Elements in the `patches` module?  If they're only used in this patch, I guess this isn't too much of an issue. Can we foresee this element being used elsewhere in the future? In that case would we move it?

troi/patches/daily_jams.py
```diff
@@ -42,13 +43,18 @@ def read(self, inputs = []):
...
+        return [ Playlist(name="Daily Jams for %s, %s" % (self.user, jam_date), recordings=days[self.day - 1]) ]
 
-class DailyJamsPatch(troi.patch.Patch):
 
-    def __init__(self, debug=False):
-        super().__init__(debug)
+class DailyJamsPatch(troi.patch.Patch):
+    """
+        Taken a list of Recordings, break them into 7 roughly equal chunks and return
+        the chunk for the given day of the week.
```
alastair, line 56: Does this return a `Playlist` with one playlist for the given day?  Is there a way that it makes sense for this to return the 7 playlists and allow us to just index into it given the day? (I've not looked too much at the implementation of this pipeline, so I'm not sure if it's possible or a good idea)

troi/patches/weekly_flashback_jams.py
```diff
@@ -0,0 +1,138 @@
...
+import troi.musicbrainz.year_lookup
+
+
+@click.group()
+def cli():
+    pass
+
+class DecadePlaylistSplitterElement(Element):
+    '''
+        Split a list of recordings that have their year attribute filled out
```
alastair, line 20: nit: "Take a list of recordings....."

troi/playlist.py
```diff
@@ -1,79 +1,148 @@
...
+            print()
+
+
+    def save(self):
+
+        if not self.playlists:
+            raise PipelineError("Playlists have not been generated yet.")
+
+        for i, playlist in enumerate(self.playlists):
+            filename = playlist.filename or "playlist %03d.jspf" % i
```
alastair, line 105: can we have no spaces in this filename, to make it easy to select from a terminal?

troi/patches/weekly_flashback_jams.py
```diff
@@ -0,0 +1,138 @@
...
+    @click.argument('user_name')
+    @click.argument('type')
+    def parse_args(**kwargs):
+        """
+        Generate weekly flashback playlists from the ListenBrainz recommended recordings.
+
+        \b
+        USER_NAME is a MusicBrainz user name that has an account on ListenBrainz.
+        TYPE is The type of daily jam. Must be 'top' or 'similar'.
+        TOKEN is the user token from the LB user into whose account you wish to post this playlist
```
alastair, line 85: not an argument

troi/playlist.py
```diff
@@ -1,79 +1,148 @@
...
+                              json=_serialize_to_jspf(playlist, created_for),
+                              headers={"Authorization": "Token " + str(token)})
+            if r.status_code != 200:
+                raise PipelineError("Cannot post playlist to ListenBrainz: HTTP code %d: %s" %
+                                    (r.status_code, r.json()["error"]))
+
+            try:
+                result = ujson.loads(r.text)
+            except ValueError as err:
+                raise PipelineError("Cannot post playlist to ListenBrainz: " + str(err))
```
alastair, line 144: At some point in time I want to revisit the use of ujson here. I think it some cases it still makes sense to use it for speed, but in something like parsing this response, `r.json()` is probably easier to write, since the response is so small. Not a huge blocker to fix

troi/patches/weekly_flashback_jams.py
```diff
@@ -0,0 +1,138 @@
...
+            random.shuffle(decades[decade])
+            playlists.append(Playlist("%ss flashback jams" % str(decade), filename="%ss_flashback_jams.jspf" % str(decade),
+                             recordings=decades[decade]))
+
+        return playlists
+
+
+class WeeklyFlashbackJams(troi.patch.Patch):
+    """
+        See below for description
```
alastair, line 68: I'm not sure about this either. There must be a better way of documenting this when we have a `parse_args` method too

troi/patches/weekly_flashback_jams.py
```diff
@@ -0,0 +1,138 @@
...
+    def outputs():
+        return [Recording]
+
+    @staticmethod
+    def slug():
+        return "weekly-flashback-jams"
+
+    @staticmethod
+    def description():
+        return "Generate a weekly playlist from the ListenBrainz recommended recordings based on decades."
```
alastair, line 105: incorrect now? We should remove this method and use the first line of `parse_args` somehow. I'll look into it

troi/playlist.py
```diff
@@ -1,79 +1,148 @@
...
 
-from troi import Recording
+from troi import Recording, Playlist, PipelineError, Element
 from troi.operations import is_homogeneous
-import troi
 
+LISTENBRAINZ_PLAYLIST_CREATE_URL = "https://test.listenbrainz.org/1/playlist/create"
 
-class PlaylistElement(troi.Element):
+def _serialize_to_jspf(playlist, created_for=None):
```
alastair, line 10: would be great to not duplicate this code between troi and LB. Any thoughts?

troi/patches/weekly_flashback_jams.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def create(self, inputs):
+        print(inputs)
```
alastair, line 108: debugging can be removed

alastair (commented):
Mostly good, a few small things.

I got quite confused with the difference between a `Playlist` (an Entity, with a name and list of recordings), and a `PlaylistElement` (an Element, with a list of playlists). I don't think that we can make this better, we just have to make sure that it's clearly documented.
Now there are enough moving parts that it might be a good idea to generate and publish the documentation (I already registered https://troi.readthedocs.io/en/latest/) so that people can see an overview of all of the moving parts.

troi/playlist.py
mayhem: I'll skip this for now, lets chat about this tomorrow.

troi/patches/weekly_flashback_jams.py
mayhem: I see no obvious hacks/magic that improves on this. Do you know anything?

troi/playlist.py
mayhem: Nothing easy, sadly.

troi/cli.py
mayhem: Done.

troi/cli.py
mayhem: done.

troi/patches/daily_jams.py
mayhem: My plan was to leave them in the patches until it was clear it would be useful in another context and then move it.

troi/playlist.py
mayhem: done

troi/playlist.py
mayhem: done

troi/patches/daily_jams.py
mayhem: Possibly. I think all of this is still in flux that I don't want to mess with it right now.

troi/patches/daily_jams.py
alastair: fine by me

troi/playlist.py
alastair: We discussed this in person, we can probably use `r.json()` in troi when we read small bits of data from an API using requests. For the bigger picture of json parsing in the LB listen pipeline we need to look at this in more detail

troi/playlist.py
alastair: as we discussed, we should look at moving this to an external python package that both troi and LB can import

troi/patches/weekly_flashback_jams.py
alastair: It depends on what we want this information for. If we want to be able to read the file and see the description in this class then there's no great way to do it other than duplicating it.
If we want this available on the sphinx documentation, we can programatically hack `WeeklyFlashbackJams.__doc__` and populate it with the contents of `parse_args`'s docs.

alastair (approved):
lgtm!

PR #34, opened 2021-01-22
Improved mbid mapping, playlist submit improvements and year lookup improvements.

mayhem:
This is the troi work that was done to improve playlists and to use the new MBID mapping that was created for the year end 2020 reports. The mapping code hasn't changed substantially, mostly copypasta. Also the year lookup has a new option for keeping items that were not found during the year lookup.


troi/webserver/main.py
```diff
@@ -32,19 +32,19 @@ def page_not_found(e):
...
     for input in inputs:
         arg = request.args.get(input['name'], '')
         if not arg and not input['optional']:
             return "Parameter %s is missing." % input['name'], ()
         try:
-            args.append(input['type'](arg))
+            args[input['name']] = input['type'](arg)
         except:
             if input['optional']:
                 args.append(None)
```
amCap1712, line 45: Since `args` is now a `dict`, we'll need to change this accordingly.

troi/playlist.py
```diff
@@ -27,7 +41,7 @@ def _serialize_to_jspf(playlist, created_for=None):
         track["identifier"] = "https://musicbrainz.org/recording/" + str(e.mbid)
         if artist_mbids:
             track["extension"] = {
-                "https://musicbrainz.org#jspf": {
+                PLAYLIST_TRACK_URI_PREFIX: {
                     "artist_mbids" : artist_mbids,
```
amCap1712, line 45: We should probably add these fields to docs at https://musicbrainz.org#jspf as well.

troi/playlist.py
```diff
@@ -1,19 +1,33 @@
...
+                     "public": True
+                 }
+           }
+    }
 
     if playlist.name:
         data["title"] = playlist.name
 
+    if playlist.description:
+        data["annotation"] = playlist.description
```
amCap1712, line 29: We should probably add this fields to docs at https://musicbrainz.org#jspf.

amCap1712 (commented):
Looks Good!

troi/webserver/main.py
mayhem: Good catch!

troi/playlist.py
mayhem: https://wiki.musicbrainz.org/jspf?diff=75246&oldid=74988 we had it but it was named wrong.

troi/playlist.py
mayhem: annotation is a field defined in jspf itself.

amCap1712 (approved):

PR #39, opened 2021-08-10
Update AnnoyElement to match the similarity api in AB prod

### LINKED ISSUE #38: 404 error on similarity.acousticbrainz.org
hello again, 

https://similarity.acousticbrainz.org/api/v1/similarity/tzanetakis/480b391b-827a-4c92-b98f-5b223b74f22c is giving 404 error but https://acousticbrainz.org/480b391b-827a-4c92-b98f-5b223b74f22c exist 

command used : `sudo python3 -m troi.cli playlist ab-similar-recordings -p --debug 480b391b-827a-4c92-b98f-5b223b74f22c tzanetakis`

amCap1712:
It appears that we have changed AB's similarity API endpoint when finalizing the annoy PR. Hence, the AnnoyElement in Troi needs to be fixed up as well. Notably, the recording_ids should now be passed as a query param and the JSON returned by the endpoint looks like the example in the updated test. This should hopefully resolve #38.

### DIFF
--- a/troi/acousticbrainz/annoy.py
+++ b/troi/acousticbrainz/annoy.py
@@ -42,7 +42,8 @@ class AnnoyLookupElement(Element):
 
         url = self.SERVER_URL + self.metric
         self.debug(f"url: {url}")
-        r = requests.get(url, params={'remove_dups': 'true', 'recording_ids': [self.mbid]})
+        # recording_ids format is mbid:offset;mbid:offset
+        r = requests.get(url, params={'remove_dups': 'true', 'recording_ids': self.mbid + ":0;"})
         if r.status_code != 200:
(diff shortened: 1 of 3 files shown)


troi/acousticbrainz/annoy.py
```diff
@@ -52,7 +52,7 @@ def read(self, inputs):
             raise PipelineError("Cannot fetch annoy similarities from AcousticBrainz: Invalid JSON returned: " + str(err))
 
         entities = []
-        for row in results:
+        for row in results[self.mbid]["0"]:
```
amCap1712, line 55: I don't think the Annoy element supports specifying custom offsets for mbids yet so we can safely hardcode the "0".

alastair: If you don't specify an offset in the query then it defaults to 0. Hard-coding here is fine

PR #45, opened 2022-02-15
Fix area random recordings

mayhem:
Quick fix to get the random area recordings working again. Still no tests. Changes include:

- Fetch artist mbids
- Fetch artist area mbid, not id
- Use dataset hoster to fetch the data, remove the old unneeded file
- Use improved PlaylistRedundancyReducerElement

### DIFF
--- a/troi/patches/area_random_recordings.py
+++ b/troi/patches/area_random_recordings.py
@@ -3,11 +3,12 @@ import datetime
 import click
 
 from troi import PipelineError, Recording
-import troi.listenbrainz.area_random_recordings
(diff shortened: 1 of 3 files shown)


amCap1712 (approved):
The area lookup failing test can be fixed by updating the mock:
https://github.com/metabrainz/troi-recommendation-playground/blob/c4844c032b8644638ab7d831afc48448d9ce1492/troi/tools/tests/test_area_lookup.py#L16
 The other failing test seems not related to this PR so fine to ignore. LGTM otherwise. Thanks!

PR #47, opened 2022-04-11
General cleanup

mayhem:
This PR does the following things:

1. Remove all components that use AB.
2. Move the YIM components to an internal folder so that they are out of the way for day-to-day sort of stuff. We really only need them for YIM times.
3. Fix the failing test for the MBID mapping


amCap1712 (approved):
May rename internal to yim but otherwise LGTM. Also, the YIM patches may still be broken but those can be fixed later. Thanks!

mayhem:
I'm sure the YIM patches are broken, since the data doesn't exist. Which is why I wanted them out of the way -- and I am not fussed what we call that dir, as long as new comers don't get caught up in it.

PR #48, opened 2022-04-16
LB-1087: Get weekly flashback jams working again

amCap1712:
LB-1087
Need to fix invalid method argument name and replace obsolete year lookup element with mbid mapping element.


mayhem (approved):
Thanks!

PR #50, opened 2022-04-17
Fix world trip

sivannavis:


sivannavis:
I think it should be fine after updating the API addresses, but it still runs into the time-out error `HTTP code 504` for some queries in http://wolf.metabrainz.org:8000/area-random-recordings.

sivannavis:
The time-out error was solved by ignoring it temporarily so that it can spit out a playlist from different countries after all.
However, it still takes too long to make a full world trip.

mayhem:
Thanks, but this PR is effectively superseded by PR #51.

PR #52, opened 2022-04-21
Add patch to save current recommendations to a collaborative playlist

mayhem:
This PR add the functionality to create a playlist from the current recommendations for use in evaluating our various collaborative filtering tests.

Also, return a non-api URL that the user who submitted the playlist can click on.


amCap1712 (approved):
LGTM, Thanks!

PR #53, opened 2022-04-22
Minor fixes to recommendation playlist feature.

mayhem:
Read the fields names we're expecting for the model_id and model_url... And one more print_recording fix.


troi/patches/daily_jams.py
```diff
@@ -111,7 +111,10 @@ def create(self, inputs, patch_args):
         artist_filter = troi.filters.ArtistCreditFilterElement([])
         artist_filter.set_sources(r_lookup)
 
-        artist_limiter = troi.filters.ArtistCreditLimiterElement()
+        if type == "top":
```
mayhem, line 114: This is an attempt do make the top artists list better... Will likely change again.

troi/listenbrainz/recs.py
```diff
@@ -50,7 +50,7 @@ def read(self, inputs = []):
                 break
 
             lb_metadata = { "model_id": recordings["payload"].get("model_id", None),
-                            "model_info_url": recordings["payload"].get("model_info_url", None) }
+                            "model_info_url": recordings["payload"].get("model_url", None) }
```
amCap1712, line 53: Dict also needs changing?

PR #54, opened 2022-04-22
Making sure that print recording and mbid_mapping are more robust

mayhem:
Just trying to have few places where it crashes, so adding more checks for when data is missing and doing the right thing instead.


troi/core.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python3
+
+import sys
+
```
amCap1712, line 4: import click missing.

amCap1712 (approved):
LGTM, Thanks!

PR #56, opened 2022-05-26
Improve daily jams

mayhem:
Before delving in and making daily-jams super fancy, I've made a simple stop-gap measure that zips together the top and similar recs, removes duplicate tracks and shuffles the result. This PR goes hand in hand with LB PR #2016. Once this is merged we should tag it and then reference that tag in the LB PR.

### DIFF
--- a/troi/patches/daily_jams.py
+++ b/troi/patches/daily_jams.py
@@ -19,11 +19,12 @@ class DailyJamsElement(Element):
         Split weekly recommended recordings into 7 sets, one for each day of the week.
     '''
 
-    def __init__(self, recs, user, day):
+    def __init__(self, recs, user, day, slug):
         Element.__init__(self)
         self.recs = recs
         self.user = user
         self.day = day
+        self.slug = slug
 
     @staticmethod
     def inputs():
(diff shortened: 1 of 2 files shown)


amCap1712 (approved):
LGTM, Thanks!

I'd suggest to run it once if not already done to confirm it works fine.

Also, raw recs are now done. So can mix that in as well if you want.

troi/patches/daily_jams.py
```diff
@@ -14,40 +16,31 @@ def cli():
...
-        num_per_day = len(recordings) // 7
-        days = [recordings[i:i + num_per_day] for i in range(0, len(recordings), num_per_day)]
-
-        jam_date = datetime.utcnow() - timedelta(days=datetime.utcnow().isoweekday() % 7)
-        jam_date += timedelta(days=self.day)
-        jam_date = jam_date.strftime("%Y-%m-%d %a")
-
-        return [ Playlist(name="Daily Jams for %s, %s" % (self.user, jam_date), recordings=days[self.day - 1]) ]
+        output = []
+        for rec in zip_longest(inputs[0], inputs[1]):
```
amCap1712, line 37: Nit: can do like for rec0, rec1 in

mayhem: Fixed.

PR #58, opened 2022-06-03
Fix serialization issue in daily jams

amCap1712:
The patch to upload the generated playlist was failing because we are the function instead of its return value to the serialization function.


mayhem (approved):
Clearly more tests are needed! Thanks for cleaning this up!

PR #59, opened 2022-06-28
Add last listened support

mayhem:
This PR updates the daily jam to use the latest_listened_at data that is now shipped with recommendations. This PR:

- Creates a new element to filter out tracks that have been listened to recently.
- Load and process latests listened at data
- Move the currently unused ZipperElement to the operations module.

### DIFF
--- a/troi/filters.py
+++ b/troi/filters.py
@@ -321,8 +321,11 @@ class LatestListenedAtFilterElement(troi.Element):
         results = []
         now = datetime.datetime.now()
         for r in recordings:
-            td = now - r.listenbrainz["latest_listened_at"]
-            if td.days > self.min_number_of_days:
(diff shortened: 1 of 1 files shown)


troi/filters.py
```diff
@@ -289,3 +290,39 @@ def read(self, inputs, debug=False):
...
+        return [Recording]
+
+    def read(self, inputs, debug=False):
+
+        recordings = inputs[0]
+
+        results = []
+        now = datetime.datetime.now()
+        for r in recordings:
+            td = now - r.listenbrainz["latest_listened_at"]
```
amCap1712, line 324: This may need to handle the latest_listened_at null case. If we pass recommendations to this, then this case won't happen anyway but maybe useful in mixed source playlists.

amCap1712 (approved):
LGTM, Thanks!

PR #62, opened 2022-07-14
 Generate install_requires from requirements.txt and cleanups

amCap1712:
When installing troi as a library, pip will look at install_requires and not requirements.txt (which is only used during local development). Often, the two diverge because we forget to update in 1 place. So generate install_requires from requirements.txt.

The troi webserver is currently not intended to be used anywhere and can be safely removed.

Finally, move test dependencies to a separate requirements file.


mayhem (approved):
Thanks!

PR #63, opened 2022-07-15
Use version ranges instead of pinning

amCap1712:
Since we generate install_requires in setup.py from requirements.txt now, the dependencies versions should be open ended so that the application using troi as a library can install the appropriate versions. Otherwise, installing dependencies will fail with a version conflict error.


alastair (approved):
this looks fine

PR #66, opened 2022-09-28
Add spotify submission support

amCap1712:
Add support for submitting playlists to spotify.

# To discuss:
1. Whether to lookup spotify track ids as an element or in the spotify submission function
2. Whether to add a key like external_uris to our JSPF extension to store a link to the spotify playlist created


troi/playlist.py
```diff
@@ -209,6 +212,51 @@ def submit(self, token, created_for=None, algorithm_metadata=None):
...
+        response = requests.post(
+            SPOTIFY_IDS_LOOKUP_URL,
+            json=[{"[recording_mbid]": recording.mbid} for recording in recordings]
+        )
+        response.raise_for_status()
+        spotify_data = response.json()
+        for recording, lookup in zip(recordings, spotify_data):
+            recording.spotify_id = lookup["spotify_track_id"]
+
+    def submit_to_spotify(self, user_id: str, token: str, is_public: bool = True, is_collaborative: bool = False):
```
amCap1712, line 227: On a tangent, I had used a TypedDict for params earlier but dropped it because according to setup.py troi support Python 3.6. I am not sure whether troi would actually run with Python 3.6 (even 3.7 or 3.8) at the point because I don't think we just tested with whatever version LB runs on which is usually newer. We should decide which versions we want to support and according update the CI to ensure that troi indeed runs with those versions.

requirements.txt
```diff
@@ -5,3 +5,4 @@ psycopg2-binary>=2.9.3
 countryinfo>=0.1.2
 pylistenbrainz@git+https://github.com/metabrainz/pylistenbrainz.git@v0.5.2
 python-dateutil>=2.8.2
+spotipy==2.20.0
```
amCap1712, line 8: Need to figure out how this interacts with spotipy fork in LB.

troi/playlist.py
amCap1712: Discussed to settle on Python 3.8

mayhem (approved):
Tedious, but I am glad to see this getting done!

troi/tools/spotify_lookup.py
```diff
@@ -0,0 +1,107 @@
...
+    for idx, spotify_ids in index.items():
+        for spotify_id in spotify_ids:
+            if spotify_id in new_tracks_ids:
+                fixed_up_items.append((idx, spotify_id))
+                break
+
+    return fixed_up_items
+
+
+def fixup_spotify_playlist(sp: spotipy.Spotify, playlist_id: str, mbid_spotify_id_idx, spotify_id_mbid_idx):
```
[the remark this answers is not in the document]
amCap1712: Indeed!

PR #69, opened 2022-10-11
Refactor generate_playlist

amCap1712:
Accept the patch object in generate playlist and move discover_patches out of generate_playlist into the cli command. These changes should make it easier to use troi as a library while still retaining all existing functionality when used as a cli.

### DIFF
--- a/troi/core.py
+++ b/troi/core.py
@@ -1,72 +1,65 @@
 #!/usr/bin/env python3
 
 import sys
+from typing import Dict
 
 import click
 
 import troi
 import troi.playlist
 import troi.utils
+from troi.patch import Patch
+
+default_patch_args = dict(
+    debug=False,
+    echo=True,
+    save=False,
+    token=None,
+    upload=False,
+    args=None,
+    created_for=None,
+    name=None,
+    desc=None,
+    min_recordings=10
+)
 
 
-def generate_playlist(patch, debug=False, echo=True, save=False, token=None, upload=False, args=None,
-                      created_for=None, name="", desc="", min_recordings=10):
+def generate_playlist(patch: Patch, pipeline_args: Dict, supplied_patch_args: Dict):
     """
     Generate a playlist using a patch
 
-    patch: The patch slug that identifies the patch to run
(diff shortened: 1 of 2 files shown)


troi/core.py
```diff
@@ -1,72 +1,65 @@
...
+    upload=False,
+    args=None,
+    created_for=None,
+    name=None,
+    desc=None,
+    min_recordings=10
+)
+
+
+def generate_playlist(patch: Patch, pipeline_args: Dict, supplied_patch_args: Dict):
```
amCap1712, line 27: I had initially made this change because I used a TypedDict for documenting the params and it felt easier but then I realised we may be unable to use TypedDict if we want to support Py 3.6 and 3.7 so fine to revert to separate args I guess if we want.

troi/core.py
```diff
@@ -1,72 +1,65 @@
...
-        "created_for": created_for, 
-        "upload": upload, 
-        "name": name,
-        "desc": desc,
-        "min_recordings": min_recordings
-    }
-
-    pipeline = patch.create(pipelineargs, patch_args)
+    patch_args = {**default_patch_args, **supplied_patch_args}
+    pipeline = patch.create(pipeline_args, patch_args)
```
amCap1712, line 51: A few more things I was wondering was if we should merge patch_args and pipeline_args together. Because a quick look at the current patches and I see that even pipeline_args are only passed to top patch and not the entire pipeline elements.

I looked at the current uses of patch_args and I only saw it being used in YIM stuff and ForLoopElement which can probably be refactored to work with a single inputs dict.

troi/core.py
amCap1712: Discussed in IRC. will open a separate PR to use TypedDict and baseline at python 3.8

troi/core.py
amCap1712: Done.

amCap1712:
Will do remaining things discussed in IRC in separate PRs for easier review.

mayhem (approved):
Thanks!

PR #70, opened 2022-10-11
Prototype click removal from core

amCap1712:
To make it simpler to use troi as a library, we intend to remove click from the core lib. Click seemingly doesn't allow adding options/arguments to a command at runtime. So here's some cursed workaround it. I have only ported one patch for now so that we can discuss if the approach is fine.

### DIFF
--- a/troi/patches/daily_jams.py
+++ b/troi/patches/daily_jams.py
@@ -36,18 +36,8 @@ class DailyJamsPatch(troi.patch.Patch):
         can be on different dates). Recommended formatting for the date is 'YYYY-MM-DD DAY_OF_WEEK'.
         """
         return [
-            {
-                "type": "argument",
-                "args": ["user_name"],
-                "kwargs": {}
-            },
-            {
-                "type": "argument",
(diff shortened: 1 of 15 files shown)


troi/patches/daily_jams.py
```diff
@@ -23,11 +23,25 @@ class DailyJamsPatch(troi.patch.Patch):
...
     @staticmethod
-    @cli.command(no_args_is_help=True)
-    @click.argument('user_name')
-    @click.argument('jam_date', required=False)
-    def parse_args(**kwargs):
-        """
+    def get_args():
+        return [
+            {
+                "type": "argument",
```
mayhem, line 29: To get a better idea as to where this is going, it would be nice to convert a patch that has more args, or perhaps document these dicts a better... I hate that we're having to duplicate such an important module such as click, but that seems to be our lot in this life...

Overall, I think this will be fine for what we need -- it seems that there is enough flexibility to add support for future more complicated command line args as we go.

mayhem (approved):

PR #71, opened 2022-10-12
Add element to load a playlist from LB

amCap1712:
Add a method to de-serialize JSPF playlist to a `troi.Playlist` entity. Also, added a dummy patch for now but we may not want to put that in LB instead of troi or maybe go entirely some other way.

### DIFF
--- a/troi/patches/playlist_from_listenbrainz.py
+++ b/troi/patches/playlist_from_listenbrainz.py
@@ -10,7 +10,7 @@ def cli():
     pass
 
 
-class ResavePatch(troi.patch.Patch):
+class TransferPlaylistPatch(troi.patch.Patch):
 
     @staticmethod
     @cli.command(no_args_is_help=True)
@@ -34,7 +34,7 @@ class ResavePatch(troi.patch.Patch):
 
     @staticmethod
     def slug():
-        return "resave-playlist"
+        return "transfer-playlist"
 
     @staticmethod
     def description():
--- a/troi/patches/playlist_from_listenbrainz.py
+++ b/troi/patches/playlist_from_listenbrainz.py
@@ -41,8 +41,5 @@ class TransferPlaylistPatch(troi.patch.Patch):
         return "Retrieve a playlist from the ListenBrainz"
 
     def create(self, inputs):
-        token = inputs.get("read_only_token")
-        if not token:
-            token = inputs.get("token")
-        playlist = PlaylistFromJSPFElement(inputs["mbid"], token)
-        return playlist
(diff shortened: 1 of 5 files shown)


troi/patches/playlist_from_listenbrainz.py
```diff
@@ -0,0 +1,47 @@
...
+from troi.playlist import PlaylistFromJSPFElement
+import troi.musicbrainz.recording_lookup
+
+
+@click.group()
+def cli():
+    pass
+
+
+class ResavePatch(troi.patch.Patch):
```
amCap1712, line 13: We may want to put this patch in LB or go some other way. Probably best to wait until troi is refactored to be a better library.

troi/playlist.py
```diff
@@ -255,6 +284,11 @@ def submit_to_spotify(self, user_id: str, token: str, is_public: bool = True, is
             result = sp.playlist_add_items(spotify_playlist["id"], spotify_track_ids)
             submitted.append((spotify_playlist["external_urls"]["spotify"], spotify_playlist["id"]))
 
+            if playlist.external_urls:
```
amCap1712, line 287: Need to decide on exact way how this should be stored in Playlist entity. We also likely want to store the unmatched tracks or least their count.

troi/patches/playlist_from_listenbrainz.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @staticmethod
+    def description():
+        return "Retrieve a playlist from the ListenBrainz"
+
+    def create(self, inputs):
+        token = inputs.get("read_only_token")
+        if not token:
+            token = inputs.get("token")
+        playlist = PlaylistFromJSPFElement(inputs["mbid"], token)
```
mayhem, line 47: could collapse into one line.

troi/playlist.py
```diff
@@ -396,3 +431,32 @@ def read(self, inputs):
...
+
+    def read(self, inputs):
+        headers = None
+        if self.token:
+            headers = {"Authorization": f"Token {self.token}"}
+        response = requests.get(LISTENBRAINZ_PLAYLIST_FETCH_URL + self.playlist_mbid, headers=headers)
+        response.raise_for_status()
+        data = response.json()
+
+        playlist = _deserialize_from_jspf(data)
```
mayhem, line 461: here too collapse into one line.

mayhem (approved):
I think the naming could be improved -- lets discuss in IRC.

troi/playlist.py
amCap1712: Done

PR #72, opened 2022-10-17
Add RecentListensTimestampLookup element

amCap1712:
We generate daily jams based on user's timezone in ListenBrainz. In many cases, these jams will get generated before yesterday's listens get imported in the Spark cluster. As a result the latest_listened_at timestamp, of the recommendations can be out of date. To fix this, retrieve all listens of the user from last 2 days and augument the latest_listened_at value of the recordings with those.

### DIFF
--- a/troi/patches/daily_jams.py
+++ b/troi/patches/daily_jams.py
@@ -15,6 +15,9 @@ def cli():
     pass
 
 
+DAYS_OF_RECENT_LISTENS_TO_EXCLUDE = 14  # Exclude tracks listened in last X days from the daily jams playlist
+
+
 class DailyJamsPatch(troi.patch.Patch):
     """
         Taken a list of Recordings, break them into 7 roughly equal chunks and return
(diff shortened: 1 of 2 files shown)


troi/patches/daily_jams.py
```diff
@@ -62,8 +61,11 @@ def create(self, inputs):
         raw_recs_lookup = troi.musicbrainz.recording_lookup.RecordingLookupElement()
         raw_recs_lookup.set_sources(raw_recs)
 
+        recent_listens_lookup = troi.listenbrainz.listens.RecentListensTimestampLookup(user_name, days=2)
+        recent_listens_lookup.set_sources(raw_recs_lookup)
+
         latest_filter = troi.filters.LatestListenedAtFilterElement(14)
```
mayhem, line 67: Make this a constant please, since this is a critical tuning value for the patch.

mayhem (approved):
One minor improvement, but ready to merge after that. Thanks!

PR #75, opened 2022-10-22
Add more tracks to daily jams if needed

amCap1712:
Users who regularly listen to daily jams start to see very less tracks in their playlists. Currently, we consider top 100 raw recs as the source of daily jams playlists and filter out any tracks listened in the past 14 days. Since recommendations don't change a lot on daily basis, most of the tracks get filtered.

To tackle this situation, we add the next 100 raw recommendations of the user to the source. Note that we do not fetch top 200 directly because that would mean an equal chance of recommending a track from 101-200 range as a track from 1-100. We don't want that, we want to prefer the top 100 over the next 100. Hence, only ask for more recordings if there is a shortfall.

### DIFF
--- a/troi/listenbrainz/listens.py
+++ b/troi/listenbrainz/listens.py
@@ -7,12 +7,15 @@ from troi import Element, Recording
 
 
 class RecentListensTimestampLookup(Element):
-    """ Element to look up the time when a user last listened given recordings in past X days """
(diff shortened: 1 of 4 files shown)


mayhem (approved):
scary. but hey. we'll replace all this soon. :see_no_evil:

PR #76, opened 2022-10-22
Create a playlist from the similar recordings of your top recordings

mayhem:
This patch uses the statistics (top recordings) and similar recordings to look up the first two most similar recordings of your top recordings. Deduplicate, lookup metadata, artist credit filter, done!

Given how low the counts on the current dataset are, the results are rather quite impressive.


troi/patches/top_recording_similar_recordings.py
```diff
@@ -0,0 +1,133 @@
...
+    @staticmethod
+    def outputs():
+        return [Recording]
+
+    def read(self, inputs):
+
+        sim = []
+        for recording in inputs[0]:
+            url = f"https://labs.api.listenbrainz.org/similar-recordings/json"
+            r = requests.get(url, { "recording_mbid": recording.mbid, "algorithm": "session_based_days_7_session_300" })
```
amCap1712, line 42: can use count = 2 to avoid the server from doing extra work since we are only using the first 2 recs.

PR #79, opened 2022-11-09
Improve support for additional playlist metadata

amCap1712:
1. Store external uris of the playlist in additional metadata.
2. Add methods to easily update additional metadata of the playlist. These are needed so that adding some metadata in a function does not override the existing data and so that we don't need to have such checks in multiple places.
3. Move DumpElement from utils to playlist to avoid circular import issues.
4. Export to spotify before uploading to LB so that the external url is uploaded to LB through additional_metadata.
5. Unrelated: fix a spotify playlist fixup when there are no alternative ids available to fix up tracks.

### DIFF
--- a/troi/utils.py
+++ b/troi/utils.py
@@ -58,11 +58,17 @@ def discover_patches_from_dir(module_path, patch_dir, add_dot=False):
 
 
 def recursively_update_dict(source, overrides):
-    """ Update a nested dictionary recursively in place. """
+    """ Updates the `source` dictionary in place and in a recursive fashion. That is unlike
+    dict1.update(dict2) which would simply replace values of keys even in case one of the
+    values is dict, this method will attempt to merge the nested dicts.
+
+    Eg: dict1 - {"a": {"b": 1}}, dict2 - {"a": {"c": 2}}
+    dict1.update(dict2) - {"a": {"c": 2}}
+    recursively_update_dict(dict1, dict2) - {"a": {"b": 1, "c": 2}}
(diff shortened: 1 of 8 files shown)


troi/utils.py
```diff
@@ -59,53 +57,13 @@ def discover_patches_from_dir(module_path, patch_dir, add_dot=False):
...
-    @staticmethod
-    def outputs():
-        return [Recording, Playlist]
-
-    def read(self, inputs):
-        for input in inputs:
-            pr = PrintRecordingList()
-            pr.print(input)
+def recursively_update_dict(source, overrides):
+    """ Update a nested dictionary recursively in place. """
```
mayhem, line 61: This docstring really doesn't tell me what this function does. What keys are affected? What do the arguments mean?

troi/utils.py
```diff
@@ -59,53 +57,13 @@ def discover_patches_from_dir(module_path, patch_dir, add_dot=False):
...
+    for key, value in overrides.items():
+        if isinstance(value, dict) and value:
+            updated = recursively_update_dict(source.get(key, {}), value)
```
mayhem, line 64: could be one line.

mayhem (approved):
Tiny improvements only, looks good!

troi/utils.py
amCap1712: Done

troi/utils.py
amCap1712: Updated doc with example

PR #80, opened 2022-11-09
Add support for updating existing spotify playlists

amCap1712:
Spotify does not allow us to delete playlists only unfollow them. So instead of creating new ones daily, we want to update the existing one over and over.


mayhem (approved):
Good solution to store the spotify playlist id in our playlist!

PR #81, opened 2022-11-21
Improve spotify playlist handling

amCap1712:
Handle case if user has deleted spotify playlist

If the spotify url provided to update existing playlist returns an error, it means that the user may have unfollowed/deleted it from spotify. In that case, attempt to create a new playlist.

Do not attempt to update existing playlist if url is None


mayhem (approved):
Yep, I deleted my playlist, expecting a new one to be created. :)

PR #82, opened 2022-11-22
Use more recommendations if needed for daily jams

amCap1712:
We may exhaust top 200 recordings for a user, in that case keep adding more recordings from ListenBrainz recommendations (we have 1000 available) till either the recordings from recommendations exhaust or we have a minimum recordings available for daily jams.

### DIFF
--- a/troi/patches/daily_jams.py
+++ b/troi/patches/daily_jams.py
@@ -76,27 +76,37 @@ class DailyJamsPatch(troi.patch.Patch):
 
         return latest_filter
 
-    def check_and_add_more_recordings(self, user_name, recordings):
+    def get_recordings(self, user_name):
         # get the list of recordings we have so far, users who regularly listen to daily jams will have
         # most of their tracks get filtered out because the recs don't change a lot on daily basis. so
         # if there is a shortfall of tracks, move on the next 100 recommendations of the user. we could
         # fetch top 200 directly in first place but then there would be an equal chance of recommending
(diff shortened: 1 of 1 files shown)


troi/patches/daily_jams.py
```diff
@@ -76,43 +76,45 @@ def apply_filters(self, user_name, element):
...
         # 100 over the next 100. so only ask for more recommendations if there is a shortfall.
-        if len(recordings) < DAILY_JAMS_MIN_RECORDINGS:
+
+        # we have top 1000 recordings available in ListenBrainz. so move on to next 100 in this same manner
+        # till we exhaust all recommendations available in ListenBrainz for the user or we have obtained the
+        # minimum of recordings we require for daily jams.
+
+        offset = 0
+        all_recs = []
+        while len(all_recs) < DAILY_JAMS_MIN_RECORDINGS and offset <= 900:
```
mayhem, line 93: Maybe define a BATCH_SIZE and then subtract that from DAILY_JAMS_MIN_RECORDINGS?

mayhem (approved):

troi/patches/daily_jams.py
amCap1712: Hmm, not sure what you mean? The number of recs one iteration of the loop adds is variable, depends on how many recordings are removed by various filters.

mayhem: I really really dislike seeing constants in code. 900 and 100 in this case. Define them all or derive them from defined values.

PR #83, opened 2022-11-22
Consider canonical recordings in latest listen filter

amCap1712:
To reduce some of the "dupes", different recordings in MB but close enough for a listener consider canonical mbids of a recordings during filtering. As discussed in IRC, this is likely not going to be enough for weeding out all undesired recordings.

### DIFF
--- a/troi/listenbrainz/listens.py
+++ b/troi/listenbrainz/listens.py
@@ -1,5 +1,6 @@
 from collections import defaultdict
 from datetime import datetime, timedelta
+from typing import Optional
 
 import requests
 
@@ -15,7 +16,7 @@ class RecentListensTimestampLookup(Element):
         super().__init__()
         self.user_name = user_name
         self.days = days
-        self.index = None
+        self.index: Optional[dict[str, int]] = None
 
     @staticmethod
     def inputs():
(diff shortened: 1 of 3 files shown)


troi/listenbrainz/listens.py
```diff
@@ -58,15 +46,44 @@ def _fetch_recent_listens_index(self):
...
+        ts1 = self.index.get(r.mbid)
+
+        # check latest listened timstamp for canonical mbid
+        ts2 = None
+        if r.listenbrainz is not None and r.listenbrainz.get("canonical_recording_mbid"):
+            canonical_mbid = r.listenbrainz.get("canonical_recording_mbid")
+            ts2 = self.index.get(canonical_mbid)
+
+        # if we have timestamp for both canonical and normal mbid take the maximum of two else whichever
+        # one is available. if none is available continue ahead
```
mayhem, line 77: what was your reasoning for choosing the larger one?

mayhem (approved):

troi/listenbrainz/listens.py
amCap1712: Larger one means that it was listened more recently. Say one mbid listened 3 days ago and the other mbid listened yesterday. Store yesterday's timestamp for use in filter later.

PR #86, opened 2022-12-22
YIM 2022 playlist touch-ups

mayhem:
Not a whole lot of changes from last year were required to get to the two popular playlists up and running for this year. The most significant change is to the top missed tracks, since the data now resides in a local DB (created by spark and imported into the DB). The top missed code that used to reside in the bono-data-sets repo is now part of this patch.

The top-discoveries patch is straightforward -- it queries the datasets.lb.org endpoints. The top missed recordings, however, requires that complete DB connect strings for the LB database containing the similar users and the MB database containing the mapping.tracks_of_the_year tables are passed on the command line.

Both patches also require arguments of user_id and user_name, for reasons previously discussed.

### DIFF
--- a/troi/patches/top_missed_recordings_for_year.py
+++ b/troi/patches/top_missed_recordings_for_year.py
@@ -40,7 +40,6 @@ class MissedRecordingsElement(Element):
                              FROM mapping.tracks_of_the_year t
                             WHERE user_id = %s
                          ) SELECT recording_mbid
-                                , recording_name
                                 , sum(listen_count) AS listen_count
                              FROM mapping.tracks_of_the_year t
                             WHERE user_id IN %s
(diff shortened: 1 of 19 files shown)


amCap1712 (approved):
LGTM, Thanks!

troi/patches/top_missed_recordings_for_year.py
```diff
@@ -0,0 +1,159 @@
...
+    def outputs():
+        return [ Recording ]
+
+    def read(self, inputs):
+
+        output = []
+        with psycopg2.connect(self.db_connect_str) as conn:
+            with conn.cursor(cursor_factory=psycopg2.extras.DictCursor) as curs:
+ 
+                query = """WITH exclude_tracks AS (
```
amCap1712, line 38: Can remove the extra fields from the query because only using recording_mbid from results

mayhem: Oh, the listen_count should be written to the LB object. Both fixed.

PR #91, opened 2023-01-26
Daily jams simplification, New Jams and general improvements

mayhem:
This PR covers a few things that have changed since I dug into the problems with tracks repeating constantly in daily jams:

- Daily jams now first processes all the raw recs (data lookup, latest listens filter, hate filter) and then randomly choose 50 tracks from those that remain. The playlist generation is slower this way -- but it makes the patch much simpler overall. Very soon some of these recs processing steps should move to Spark as much as possible. But now Daily Jams will pick tracks from the entire range of tracks available, not just the top tracks.
-  The "never listened to this track" feature wasn't implemented (there was a comment for it, but the code wasn't), but there is an element for filtering out/keeping the unlistened tracks.
- This new element also begat the now super simple New Jams patch, which just plucks the unlistened tracks, remove hated tracks and returns that as a playlist.
- The artist redundancy and shuffle features have also been incorporated into the PlaylistMakerElement, which simplifies patches quite a bit.
- The Recording lookup element has been extended to handle Playlists as well as Recordings. Something I thought would be needed, but wasn't. Keeping it anyway since it may be useful later.


troi/patches/new_jams.py
```diff
@@ -0,0 +1,80 @@
...
+        return "new-jams"
+
+    @staticmethod
+    def description():
+        return "Generate a playlist from the ListenBrainz recommended recordings that the user has not listened to."
+
+    def create(self, inputs):
+        user_name = inputs['user_name']
+
+        recs = troi.listenbrainz.recs.UserRecordingRecommendationsElement(user_name, "raw", count=1000)
```
amCap1712, line 53: Can you test whether pylistenbrainz automatically converts this into batch of calls because IIRC LB API will return 100 at max in one call so either pylb or a loop here has to batch calls accordingly.

amCap1712 (approved):
LGMT Thanks! I'll suggest to a sanity check run before merging though.

troi/patches/new_jams.py
mayhem: 1000 tracks are actually returned, yes. But I will triple check before merging.

mayhem: It isn't pylistenbrainz that does, but datasethoster that does:

https://github.com/metabrainz/troi-recommendation-playground/blob/main/troi/listenbrainz/recs.py#L37

And I've been listening to the new daily jams for 5 days now, solid, with much better results.

PR #92, opened 2023-02-11
Make the date a pass-through value

mayhem:
We no longer need to parse the jam date -- we're just naming the playlist with it, so just pass it through.


amCap1712 (approved):
LGTM, Thanks!

PR #95, opened 2023-03-01
Data fetching rate limiting

mayhem:
We suspect that Daily Jams are failing for some users because we're not sufficiently handling rate limiting in our patches. This PR aims to fix that, so we can get daily jams going again.


amCap1712 (approved):
LGTM, Thanks

PR #96, opened 2023-03-04
Use an auth token for extended ratelimit

amCap1712:
Daily jams are failing for users in production due to ratelimit errors. The intent is to pass troi's auth token to the patch in prod so that the ratelimit can be bypassed.

### DIFF
--- a/troi/core.py
+++ b/troi/core.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-import http.client
 import sys
 from typing import Dict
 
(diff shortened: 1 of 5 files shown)


amCap1712:
tested in prod. this still doesn't scale to all users.

amCap1712:
There was a bug in getting the token due to which we never got the extended ratelimit earlier. fixed now.

mayhem:
I'm not fond of having to pass around the auth token, but this works fine for now. Before too long I want to implement Patch Local Storage, so that auth keys, playlist names are whatever else that the entire pipeline might be interested in using is available to all elements without passing around so many things.

amCap1712:
That Patch Local Storage sounds good to me.

PR #98, opened 2023-05-08
Turn daily-jams patch into periodic-jams

mayhem:
The code for daily-jams and weekly-jams is nearly identical and weekly-new-jams is only one slight change, save for the name of the playlist. This PR consolidates all these features into the periodic-jams.

There is a new argument to this patch: type. It must be one of daily-jams, weekly-jams and weekly-new-jams. If no arguments are given, daily-jams are created.


troi/patches/periodic_jams.py
```diff
@@ -15,27 +15,35 @@
...
 
         \b
         USER_NAME is a MusicBrainz user name that has an account on ListenBrainz.
+        TYPE Must be one of "daily-jams", "weekly-jams" or "weekly-new-jams".
         JAM_DATE is the date for which the jam is created (this is needed to account for the fact different timezones
         can be on different dates). Required formatting for the date is 'YYYY-MM-DD'.
         """
         return [
             {"type": "argument", "args": ["user_name"]},
+            {"type": "argument", "args": ["type"], "kwargs": {"required": False}},
```
amCap1712, line 46: I see the shortcoming of the way we generate commands now. We can't pass a click.Choice here without importing click. :disappointed:

troi/patches/periodic_jams.py
```diff
@@ -45,26 +53,47 @@ def outputs():
...
+            never_listened = troi.filters.NeverListenedFilterElement()
+            never_listened.set_sources(recent_listens_lookup)
+            if jam_type == "daily-jams":
+                jam_name = "Daily Jams"
+            else:
+                jam_name = "Weekly Jams"
+                jam_date = "week of " + jam_date
+        elif jam_type == "weekly-new-jams":
+            # Remove tracks that have been listened to before.
+            never_listened = troi.filters.NeverListenedFilterElement(remove_unlistened=False)
```
amCap1712, line 91: The NeverListenedFilterElement could probably be renamed to something else and the name of the parameter as well but I can't think of a good alternative so fine to do later I guess.

amCap1712 (approved):
LGTM, Thanks!

PR #99, opened 2023-05-31
Fix periodic patch slug

mayhem:
In the previous PR the slug was set to period-jams, which doesn't allow us to distinguish between weekly-jams, daily-jams and weekly-new-jams. Setting the slug according to type fixes this.


amCap1712 (approved):
LGTM, Thanks!

PR #102, opened 2023-06-14
Add .readthedocs.yaml

mayhem:
read the docs warns us that we need to update our config:
  
   https://blog.readthedocs.com/migrate-configuration-v2/

This PR copies the config file used for lb.


amCap1712 (approved):
LGTM, Thanks

PR #103, opened 2023-06-28
Add support for expires_at in additional_metadata and add descriptios for periodic playlists

mayhem:
The bulk generation of playlists doesn't have good playlist descriptions, so this PR adds those to the troi patch, for easy reference from the listenbrainz codebase.

Also in this PR is support for expires_at in playlist's additional_metadata field, so that we know when a playlist is due to expire and can clearly show a timer icon to our users.


mayhem:
And for innocent formatting changes snuck in too. :(

amCap1712 (approved):
LGTM, Thanks

PR #105, opened 2023-07-20
LB Radio

mayhem:
This PR implements LB Radio!

The PR creates the following things:
- A prompt parser that parses a given prompt into full parse data.
- A LB Radio patch.
- Several elements that comprise the LB Radio patch, including one class for artist, tag, collection and playlist.
- An element to weigh and blend the created radio streams.
- Local Storage for patches to store and share data between elements
- Support for generating the playlist name and description and storing them into local_storage["_playlist_name"] and local_storage["_playlist_desc"] and having the PlaylistMakerElement pick them up.
- Support for post_processing steps after the patch has been run. This allows better playlist names/desc to be created.
- Support for user feedback (stored in local_storage["user_feedback"]) as a list of strings that will be shown to the user once the playlist is generated. This is useful for giving feedback about the generation process and why it may not have generated a good playlist.
-  Support for DataSetSplitter (not great, but might be improved later) and plists (in use currently)

LB Radio isn't done, nor is this code fully debugged or perfect. But, at this time we need more community feedback and before we do that, I would like to merge this PR to avoid it getting any bigger.  However, the code is working enough to be sure that merging it isn't going to create any issues for our production systems.


troi/core.py
```diff
@@ -48,6 +48,8 @@ def generate_playlist(patch: Patch, args: Dict):
 
     patch_args = {**default_patch_args, **args}
     pipeline = patch.create(patch_args)
+    _set_element_patch(patch, pipeline)
+    patch.local_storage["__test"] = "foo"
```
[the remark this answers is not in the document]
mayhem: No, removed.

troi/listenbrainz/dataset_fetcher.py
```diff
@@ -13,62 +13,99 @@ class DataSetFetcherElement(Element):
...
-                else:
-                    artist.mbids = row['artist_mbids']
+        return r
 
-            r.artist = artist
+    def create_artist(self, row):
 
-            if 'recording_name' in row:
-                r.name = row['recording_name']
+        if 'artist_mbid' in row:
```
amCap1712, line 71: Instead of returning Artist directly, maybe create an Artist object and try to set all items in it. Also, reuse this method in create_recording?

troi/listenbrainz/dataset_fetcher.py
```diff
@@ -13,62 +13,99 @@ class DataSetFetcherElement(Element):
...
+            data = r.json()[3]["data"]
+        except (IndexError, KeyError):
+            data = r.json()
+
+        output = []
+        for row in data:
+            if "recordind_mbid" in row:
+                r = self.create_recording(row)
+            elif "artist_mbid" in row:
+                r = self.create_artist(row)
```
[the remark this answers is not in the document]
mayhem: It is and was planned for, but hasn't happened so far.

troi/parse_prompt.py
```diff
@@ -0,0 +1,113 @@
...
+import pyparsing.exceptions
+
+OPTIONS = ["easy", "hard", "medium", "and", "or", "nosim"]
+
+
+class ParseError(Exception):
+    pass
+
+
+def build_parser():
```
amCap1712, line 14: I'll have to read pyparsing docs to understand how this works but if this works then sounds fine to me.

troi/listenbrainz/dataset_fetcher.py
mayhem: I dont understand what you're saying in the first sentence. Please elaborate.

amCap1712: I mean do it the the same way as in `create_recording`, `a = Artist()` then each if checks and sets the relevant field of the artist object.

mayhem: I tried that, but the code ends up being more complicated that way, since there is a possibility of returning None, rather than an empty Artist object. I prefer to keep as is.

amCap1712 (approved):
LGTM, Thanks.

PR #106, opened 2023-08-04
Add stats, collection and recs entities, improve docs, break out classes

mayhem:
This PR generally improves on the LB Radio feature, breaking classes into more modules and creating a sub-dir for the lb-radio classes. Documentation was also slightly improved and finally, stats and recs entities are now supported. The syntax also had some improvements, paving the way for knowing the user name and also allowing weights to be optional and even user names for the stats term.

The entity, collection and recs entities are new code and need to be reviewed -- all the other entities were moved, so they do not need a full review.


amCap1712 (approved):
LGTM, Thanks

PR #108, opened 2023-08-10
Prompt parsing fixes

mayhem:
Putting LB radio in front of a speaker of a non-latin based language showed a number of issues right off the bat. lol.

This PR:
- Allows for spaces in artist names.
- Allows latin characters in artist names outside of parens. Unicode only in parens.
- Fetch only 6 similar artists for the artist entity
- Update docs to explain syntax a bit.


amCap1712 (approved):
LGTM, Thanks!

PR #109, opened 2023-08-11
Improve error handling for LB-radio

mayhem:
This PR improves a number of things:

- tags can now contain all latin characters if no parens are used. (e.g. tag:blümchen)
- tags can now contain unicode characters, if parens are used. (e.g. tag:(モーニング娘)
- use new popular recordings endpoint
- Use the metadata provided in the popular recordings endpoint
- Make recording lookup optional if no recordings need looking up.
- Implement common error detection and give better feedback.

### DIFF
--- a/troi/musicbrainz/recording_lookup.py
+++ b/troi/musicbrainz/recording_lookup.py
@@ -38,7 +38,7 @@ class RecordingLookupElement(Element):
 
         data = []
         for r in recordings:
-            if r.artist.name == "" or len(r.artist.mbids) == 0 or r.name == "":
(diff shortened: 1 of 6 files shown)


troi/musicbrainz/recording_lookup.py
```diff
@@ -38,7 +38,12 @@ def read(self, inputs):
 
         data = []
         for r in recordings:
-            data.append({ '[recording_mbid]': r.mbid })
+            if r.artist is None or r.artist.name == "" or len(r.artist.mbids) == 0 or r.name == "":
```
amCap1712, line 41: This check seems wrong.

PR #110, opened 2023-08-11
Fix the stats that are not fetching metadata

mayhem:
Was checking for "", should be None.

Also, remove debug prints

### DIFF
--- a/troi/musicbrainz/recording_lookup.py
+++ b/troi/musicbrainz/recording_lookup.py
@@ -38,7 +38,7 @@ class RecordingLookupElement(Element):
 
         data = []
         for r in recordings:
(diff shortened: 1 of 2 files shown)


troi/musicbrainz/recording_lookup.py
```diff
@@ -38,7 +38,7 @@ def read(self, inputs):
 
         data = []
         for r in recordings:
-            if r.artist is None or r.artist.name == "" or len(r.artist.mbids) == 0 or r.name == "":
+            if r.artist is None or r.artist.name == None or len(r.artist.mbids) == 0 or r.name == None:
```
amCap1712, line 41: nit: is None

amCap1712 (approved):
LGTM, Thanks

MonkeyDo (approved):
See, python feels just like javascript sometimes :)

👍

PR #111, opened 2023-09-14
Improve tag radio, add troi.py, use similar tags, other tag improvements

mayhem:
This PR improves the functionality of rag radio using similar tags for medium and hard mode. The addition of similar tags can be overridden with the nosim option. Other smaller tweaks to how the tag radio works were made (adjusting percent ranges, requiring minimum tag counts).

Finally add a troi.py file, which feels like it has been missing for a while.

The similar tags dataset is currently hosted on wolf, so that will need its own PR and release into labs API; once that is done,  I will open a new PR to start using the new production data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ below.
 virtualenv -p python3 .ve
 source .ve/bin/activate
 pip3 install -r requirements.txt -r requirements_test.txt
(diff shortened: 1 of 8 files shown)


amCap1712 (approved):
LGTM, Thanks!

PR #113, opened 2023-10-25
Do not emit an album key in JSPF if we don't have album info.

mayhem:
If the key exists, but is empty LB will refuse to accept that as a listen. So not emitting the album key when we dont have an album fixes this.

### DIFF
--- a/troi/playlist.py
+++ b/troi/playlist.py
@@ -59,7 +59,8 @@ def _serialize_to_jspf(playlist, created_for=None, track_count=None):
             artist_mbids = [str(mbid) for mbid in e.artist.mbids or []]
             track["creator"] = e.artist.name if e.artist else ""
 
-        track["album"] = e.release.name if e.release else ""
+        if e.release is not None and e.release.name != "":
+            track["album"] = e.release.name
         track["title"] = e.name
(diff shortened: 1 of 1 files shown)


troi/playlist.py
```diff
@@ -68,6 +67,15 @@ def _serialize_to_jspf(playlist, created_for=None, track_count=None):
                     "artist_identifiers": artist_mbids,
                 }
             }
+
+        if e.release is not None:
+            if e.release.name != "":
+                track["album"] = e.release.name
+
+            if e.release.mbid != "":
+                track["extension"]["album"][PLAYLIST_TRACK_EXTENSION_URI]["release_identifier"] = \
```
MonkeyDo, line 76: This doesn't look right.
Should it be this?
```suggestion
                track["extension"][PLAYLIST_TRACK_EXTENSION_URI]["release_identifier"] = \
```

mayhem: That might explain why it wasn't working.  I ended up giving up for the day, lol.

MonkeyDo (approved):
Results look good on beta.LB (testing with LB Radio tracks which first shined a light on the lack of release info)

PR #114, opened 2023-10-25
add day_to_exclude var into periodic jam patch

petitminion:
I don't have a tresting env so  I didn't check if this works. Maybe we should create tests for patches ?


mayhem:
Yes, we need to create tests for patches. So much to do!

PR #115, opened 2023-11-14
LB-1395: Fix JSPF track identifier to be a list of strings

phw:
According to the XSPF/JSPF spec [1] a track can have an arbitrary number of identifiers. For JSPF this means the "identifier" field must be a list [2].

[1] https://xspf.org/spec#4112141112-identifier
[2] https://xspf.org/jspf

TODO:
- listenbrainz-server relies on the track identifier being a single string at multiple places
- Update the docs in https://musicbrainz.org/doc/jspf#track

### DIFF
--- a/troi/playlist.py
+++ b/troi/playlist.py
@@ -60,7 +60,7 @@ def _serialize_to_jspf(playlist, created_for=None, track_count=None):
             track["creator"] = e.artist.name if e.artist else ""
 
         track["title"] = e.name
-        track["identifier"] = "https://musicbrainz.org/recording/" + str(e.mbid)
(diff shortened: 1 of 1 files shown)


phw:
@mayhem @amCap1712 before I proceed here, can you verify that this format change is the proper thing to do?

mayhem:
Indeed, well spotted. However we need to take care when releasing this PR -- we need to update listenbrainz-server as well, otherwise we'll get import failures if one if releases before the other.

mayhem (approved):
LGTM, but before we merge this we need to do the LB side of things as well

PR #116, opened 2023-12-19
Update requirements.txt

petitminion:


petitminion:
to avoid dependency conflicts in fw side the future it might get usefull to use the pypi package for pylistenbrainz to allow newer version to be used ?

mayhem (approved):

georgkrause:
@petitminion TBH, if we really want to avoid dependency conflicts, troi shouldn't pin the package but specify a dependency range. Is this possible (cc @mayhem )

mayhem:
Sure, makes sense to me. Adjust the PR for that, please.

PR #117, opened 2024-01-09
Update requirements.txt

petitminion:


georgkrause:
@petitminion Are you sure this is allowed? :o

PR #118, opened 2024-01-09
Update requirements.txt

petitminion:


georgkrause:
@petitminion This will be obsolete if #119 gets merged.

PR #119, opened 2024-01-10
feat: Migrate metadata and dependencies to pyproject.toml

georgkrause:
Todos:

- [x] Documentation dependencies
- [x] Bump liblistenbrainz to properly released version

This PR includes:

- Migration to pyproject.toml
- Migration to liblistenbrainz
- Run tests on all supported python versions
- Pipeline to automatically publish releases on pypi


README.md
```diff
@@ -47,7 +47,7 @@ python3 troi.py --help
 ```
 virtualenv -p python .ve
 .ve\Scripts\activate.bat
-pip install -r requirements.txt -r requirements_test.txt
+pip install .[tests]
 python troi.py --help
```
georgkrause, line 51: We could now also run just `troi --help`, because the entrypoint enables us to do so.

@mayhem Should I update this? Do we want to remove the custom entrypoint?

georgkrause: Same thing is up to discuss for the tests below

pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+  'pyparsing'
+]
+
+[project.scripts]
+troi = 'troi.li:cli'
+
+[project.optional-dependencies]
+tests = [
+  'pytest == 7.1.2',
+  'requests-mock'
```
georgkrause, line 34: I'd always pin those dependencies, however I didn't wanted to change this without notice

pyproject.toml
```diff
@@ -0,0 +1,50 @@
+[project]
+name = "troi"
+authors = [
+  { name = "MetaBrainz Foundation", email="support@metabrainz.org"}
+]
+description="An empathetic music recommendation system pipeline"
+readme = "README.md"
+requires-python=">=3.8"
```
georgkrause, line 8: Updates the minimum python version to 3.8

georgkrause:
I added the dependencies for the documentation to pyproject.toml and updated the documentation for building the docs. I did a local test on Linux and it works as far as I can tell.

@mayhem This is ready for a review, please note the comments I left

pyproject.toml
mayhem: Fine with me, but why do you suggest this?

README.md
mayhem: Yes, lets do it.

pyproject.toml
georgkrause: @mayhem I just don't like messing around with tests that fail because a new version of a library breaks. Please note that not adding any version constraint means any currently found version might be installed.

The best scenario for me is pinning the dependencies and enabling the renovate bot in order to get a PR with a version bump, which then runs all tests so we can verify this works.

mayhem (approved):
Looks great -- thanks for all this help!

PR #122, opened 2024-01-24
LB Radio Local

mayhem:
This PR extends troi in the following way:

- Create a registry for services that allow local resolution of content: artist_search and tag_search. By doing a local search against a local collection, we avoid the problems of having to resolve a playlist, which can often destroy that playlist. This required some refactoring of the artist and tag elements.
- Refactor how playlists are generated to be a little more streamlined.
- Add support for subsonic_ids in the musicbrainz metadata.
- Simplify the syntax of LB-Radio (disallow text not wrapped in () )
- Minor other bits of refactoring.

### DIFF
--- a/troi/core.py
+++ b/troi/core.py
@@ -9,123 +9,6 @@ import troi.playlist
 import troi.utils
 from troi.patch import Patch
 
-default_patch_args = dict(
-    debug=False,
-    echo=True,
-    save=False,
-    token=None,
-    upload=False,
-    args=None,
-    created_for=None,
-    name=None,
-    desc=None,
-    min_recordings=10,
-    spotify=None
-)
-
-
-def generate_playlist(patch: Patch, args: Dict):
-    """
-    Generate a playlist using a patch
-
-    The args parameter is a dict and may containt the following keys:
-
-    * debug: Print debug information or not
-    * print: This option causes the generated playlist to be printed to stdout.
-    * save: The save option causes the generated playlist to be saved to disk.
-    * token: Auth token to use when using the LB API. Required for submitting playlists to the server. See https://listenbrainz.org/profile to get your user token.
-    * upload: Whether or not to submit the finished playlist to the LB server. Token must be set for this to work.
-    * created-for: If this option is specified, it must give a valid user name and the TOKEN argument must specify a user who is whitelisted as a playlist bot at listenbrainz.org .
-    * name: Override the algorithms that generate a playlist name and use this name instead.
-    * desc: Override the algorithms that generate a playlist description and use this description instead.
-    * min-recordings: The minimum number of recordings that must be present in a playlist to consider it complete. If it doesn't have sufficient numbers of tracks, ignore the playlist and don't submit it. Default: Off, a playlist with at least one track will be considere complete.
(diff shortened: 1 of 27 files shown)


troi/core.py
```diff
@@ -9,123 +9,6 @@
...
-    args=None,
-    created_for=None,
-    name=None,
-    desc=None,
-    min_recordings=10,
-    spotify=None
-)
-
-
-def generate_playlist(patch: Patch, args: Dict):
```
amCap1712, line 27: We use generate_playlist in LB so need to carefully update it next time we update troi there.

troi/patch.py
```diff
@@ -15,6 +31,15 @@ def __init__(self, debug=False):
         # Dict used for local storage
         self.local_storage = {}
 
+        self.patch_args = {**default_patch_args, **args}
+        self.pipeline = self.create(self.patch_args)
+        self._set_element_patch(self.pipeline)
+
+        # Setup extensible services
+        self.services = {}
+        self.register_service(RecordingSearchByTagService())
```
amCap1712, line 40: Would probably make sense to do it in LB local or other projects that expect to use it instead.

troi/patches/lb_radio_classes/artist.py
```diff
@@ -155,24 +139,28 @@ def read(self, entities):
...
                     f"Artist {artist['name']} only has {'no' if len(recs_plist) == 0 else 'few'} top recordings.")
 
-            recordings = []
-            for recording in recs_plist.random_item(start, stop, self.max_top_recordings_per_artist):
-                recordings.append(recording)
+            recordings = recs_plist.random_item(start, stop, self.max_top_recordings_per_artist)
 
             # Now tuck away the data for caching and interleaving
-            self.data_cache[artist["mbid"] + "_top_recordings"] = recordings
+            #self.data_cache[artist["mbid"] + "_top_recordings"] = recordings
```
amCap1712, line 163: intentional?

troi/playlist.py
```diff
@@ -76,6 +86,14 @@ def _serialize_to_jspf(playlist, created_for=None, track_count=None):
                 track["extension"][PLAYLIST_TRACK_EXTENSION_URI]["release_identifier"] = \
                       PLAYLIST_RELEASE_URI_PREFIX + e.release.mbid
 
+        # Output subsonic_ids to the playlist
+        subsonic_id = e.musicbrainz.get("subsonic_id", None)
```
amCap1712, line 90: Probably add to JSPF docs on MB too

amCap1712 (approved):
Didn't try out myself but looks good. Thanks!

troi/patches/lb_radio_classes/artist.py
mayhem: Nope, will remove.

troi/patch.py
mayhem: The global tag and artists searches used by LB Radio have been converted to these services, so that yes, local versions can substitute their own version, but the main, global version also needs a search defined.

troi/patches/lb_radio_classes/artist.py
mayhem: Left for a future release. The current release goal is to get Funkwhale moving again, but next week I will make another pass over troi and then work on a new release for lb-server.

troi/playlist.py
mayhem: Thanks for the reminder, done.

PR #125, opened 2024-01-29
Replace parser

mayhem:
The PyParsing parser was pretty powerful, but it had the worst error messages on the planet. By simplifying the syntax of LB-Radio a bit, the parser becomes a lot simpler and allows the super simple artist and tag queries that most people will use. This new parser is written in straight python and has much improved error messages, but I am sure that I've missed edge cases for things that need to be detected, but for now is is ok to merge. Before we release this into prod, we'll need to do a round of testing on beta to ensure that all is ok.


docs/lb_radio.rst
```diff
@@ -86,76 +86,87 @@ Syntax Notes
 ------------
 
 The syntax attempts to be intuitive and simple, but it does have some limitations. The artist: entity has the most tricky restrictions
-because it should accept the full name of an artist, so it must be wrapped in ():
+because it should accept the full name of an artist. For latin character sets, the short form for an artist name can be used:
+
+::
+
+  artist:(Blümchen)
+  artist:(The Knife)
```
amCap1712, line 94: This example should be using the short form I guess?

docs/lb_radio.rst
```diff
@@ -86,76 +86,87 @@ Syntax Notes
...
 
 ::
 
-  #(rock,pop)
-  tag:(rock,pop)
+  tag:(trip hop)
 
-These two prompts are equal, the # notation is simply a shortcut for tag. This prompt generates a playlist with recordings that have been tagged
-with both the "rock" AND "pop" tags.
+Tags that have a space in them must be enclosed in (). Specifying multiple tags requires the tags to be enclosed in () as well as comma separated.
```
amCap1712, line 157: Are tag names allowed to have comma in names in MB?

amCap1712: But I see this is same as the existing syntax so should be fine.

docs/lb_radio.rst
mayhem: Fixed.

amCap1712 (approved):
LGTM, Thanks!

PR #126, opened 2024-01-30
Read subsonic ids and locations from jspf

mayhem:
Parsing the values already written to JSPF was forgotten in previous PR, so adding now.


amCap1712 (approved):
LGTM, Thanks!

PR #128, opened 2024-02-09
Make troi quiet

mayhem:
Rather than having an --echo option, be more like other projects and have a --quiet option that can be used in server environments.


amCap1712:
@mayhem would be great if you could test the local patches and content resolver stuff locally, I don't have subsonic setup. Other than that the PR is ready.

mayhem:
Everything checks out, subsonic features working fine. Thanks!

PR #135, opened 2024-04-25
Integration fixes

mayhem:
After the refactoring of LB Radio and the implementation of the new LB Radio backends, a lot of cleanup and integration was still necessary. In this PR I:

- Finish debugging the new parser with better error messages
- Improve error message handling overall.
- Add support for the country element.
- Integrate new artist and tag search back-ends
- Get rid of old splitter class.
- Update docs to reflect new syntax.


amCap1712 (approved):
LGTM, Thanks!

PR #137, opened 2024-05-02
Adapt to new labs api query format

amCap1712:


mayhem (approved):
Working without problems.

PR #143, opened 2024-07-10
Import Playlists from SoundCloud

rimma-kubanova:
- Added SoundCloud playlist import support to `ImportPlaylistPatch`.
- Optimized code for Apple and Spotify playlist imports by removing unnecessary functions like `convert_track_to_json`.

NOTE: SoundCloud tracks store categories differently from Apple/Spotify as they do not store artist names separately. For starters, I implemented a small code snippet that splits the title into the artist name and track name. This code handles most SoundCloud tracks but does not cover all cases. We plan to implement a new function to identify SoundCloud title patterns and optimize the process for a broader set of tracks.

### DIFF
--- a/troi/tools/apple_lookup.py
+++ b/troi/tools/apple_lookup.py
@@ -1,29 +1,25 @@
 import requests
-from ratelimit import limits, sleep_and_retry
+from .utils import create_http_session
 
 APPLE_MUSIC_URL = f"https://api.music.apple.com/"
-CALLS=5
-RATE_LIMIT=1
 
-
-@sleep_and_retry
-@limits(calls=CALLS, period=RATE_LIMIT)
 def get_tracks_from_apple_playlist(developer_token, user_token, playlist_id):
     """ Get tracks from the Apple Music playlist.
     """
+    http = create_http_session()
+
     headers = {
         "Authorization": f"Bearer {developer_token}",
         "Music-User-Token": user_token
     }
-    response = requests.get(APPLE_MUSIC_URL+f"v1/me/library/playlists/{playlist_id}?include=tracks", headers=headers)
-    if response.status_code == 200:
-        response = response.json()
-        tracks = response["data"][0]["relationships"]["tracks"]["data"]
-        name = response["data"][0]["attributes"]["name"]
-        description = response["data"][0]["attributes"].get("description", {}).get("standard", "")
-    else:
-        response.raise_for_status()
-  
+    response = http.get(APPLE_MUSIC_URL+f"v1/me/library/playlists/{playlist_id}?include=tracks", headers=headers)
+    response.raise_for_status()
+
+    response = response.json()
+    tracks = response["data"][0]["relationships"]["tracks"]["data"]
+    name = response["data"][0]["attributes"]["name"]
+    description = response["data"][0]["attributes"].get("description", {}).get("standard", "")
+
     mapped_tracks = [
         {
             "recording_name": track['attributes']['name'],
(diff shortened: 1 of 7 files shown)


troi/tools/apple_lookup.py
```diff
@@ -25,7 +14,16 @@ def get_tracks_from_apple_playlist(developer_token, user_token, playlist_id):
         response = response.json()
```
mayhem, line 14: Does the apple music API have any sort of rate-limiting we might need to worry about here?

troi/tools/soundcloud_lookup.py
```diff
@@ -0,0 +1,28 @@
...
+
+SOUNDCLOUD_URL = f"https://api.soundcloud.com/"
+
+def get_tracks_from_soundcloud_playlist(developer_token, playlist_id):
+    """ Get tracks from the Apple Music playlist.
+    """
+    headers = {
+        "Authorization": f"Bearer {developer_token}",
+    }
+    response = requests.get(SOUNDCLOUD_URL+f"/playlists/{playlist_id}", headers=headers)
```
mayhem, line 11: Same question here, does SC have a rate-limit?

mayhem (commented):
Overall looks good to me, but I'm not super up on SC/Apple Music.

troi/tools/apple_lookup.py
```diff
@@ -1,7 +1,13 @@
 import requests
+from ratelimit import limits, sleep_and_retry
 
 APPLE_MUSIC_URL = f"https://api.music.apple.com/"
+CALLS=5
```
mayhem, line 5: CALLS isn't a good name -- I can't quite deduce what it means from its name. Also, I really like to note units in the commends; what is the unit of RATE_LIMIT?

amCap1712:
@rimma-kubanova Hi! This is not the right way to ratelimit API calls to Soundcloud and Apple Music. IIRC, this will actually ratelimit the number of calls of the decorated python function. What we instead want to do is make API calls to Apple Music/SoundCloud API and retry with an appropriate delay in case of ratelimit errors.

You can check this example for how to do it correctly: https://github.com/metabrainz/listenbrainz-server/blob/7ed01c43131fd1feeff6677324c9ae1b47eb06bf/listenbrainz/metadata_cache/apple/client.py#L17-L31

troi/tools/soundcloud_lookup.py
```diff
@@ -0,0 +1,34 @@
...
+@sleep_and_retry
+@limits(calls=CALLS, period=RATE_LIMIT)
+def get_tracks_from_soundcloud_playlist(developer_token, playlist_id):
+    """ Get tracks from the Apple Music playlist.
+    """
+    headers = {
+        "Authorization": f"Bearer {developer_token}",
+    }
+    response = requests.get(SOUNDCLOUD_URL+f"/playlists/{playlist_id}", headers=headers)
+    if response.status_code == 200:
```
amCap1712, line 18: The if is redundant, you can do `response.raise_for_status()` first and then the contents of this if and remove the if/else statement.

rimma-kubanova:
@amCap1712 Apologies for the misunderstanding! I've pushed a new commit implementing the example you provided and also fixed the raise_for_status part

troi/tools/apple_lookup.py
amCap1712: removed altogether now.

amCap1712:
I looked into report action and it is broken because Github (now) does not grant write access to pull requests from forks. We will have to find a replacement that works with forks as well but the PR is fine to merge. Thanks!

PR #145, opened 2024-07-24
Exporting playlist to Apple Music

rimma-kubanova:
I added the necessary functions for exporting to Apple Music. To keep the code organized, I created a class for Apple API calls in the utils folder. 

- [x] Created AppleMusicAPI and AppleException classes to handle API calls
- [x] Added submission to Apple Music in the generate_playlist function
- [x] Implemented playlist creation for Apple Music export in troi

- [x] Test Apple Music export functionality
- [x] Handle Apple lookup function fixup (Since the Apple Music API doesn't allow editing or deleting playlists, I left the lookup function in comments for future implementation.)

### DIFF
--- a/troi/__init__.py
+++ b/troi/__init__.py
@@ -311,6 +311,7 @@ class Recording(Entity):
                  ranking=None,
                  year=None,
                  spotify_id=None,
+                 apple_music_id=None,
                  musicbrainz=None,
                  listenbrainz=None,
                  acousticbrainz=None):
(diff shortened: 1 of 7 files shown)


rimma-kubanova:
**Updated PR**
* The API now successfully exports playlists to Apple Music.
* Cli has been added for easier testing
* Important Note: Due to limitations of the Apple Music API, which does not support updating or modifying existing playlists, the `fixup_apple_music_playlist ` function has not been implemented.
* Regarding existing_url: Since Apple Music only allows adding new tracks to an existing playlist, the current implementation supports appending new tracks to an existing playlist.

amCap1712 (changes requested):
Apple music does not removing tracks from a playlist either :/. I think the best way here would be to always create a new playlist. You can keep the `--apple-music-url` parameter around but add a comment to its description that its unused due to lack of API support in apple music playlists. And also update the rest of the code to always create a new playlists.

PR #147, opened 2024-08-06
SoundCloud Export & Lookup

rimma-kubanova:
PR Status: In Progress

- Developed a SoundCloud API class to manage SoundCloud interactions.
- Implemented SoundCloud lookup features, including:
   - Fixes for unplayable tracks.
   - SoundCloud Labs API ID lookup.
   - Submission functionality to SoundCloud.

- [x] test lookup functionality
- [x] test rate limits

### DIFF
--- a/troi/tools/utils.py
+++ b/troi/tools/utils.py
@@ -1,6 +1,5 @@
 import requests
 import json
-import logging
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
 
@@ -62,22 +61,6 @@ class AppleMusicAPI:
         response = self.session.get(url, headers=self.headers)
         return response.json()
 
-def create_http_session():
-    """ Create an HTTP session with retry strategy for handling rate limits and server errors.
-    """
-    retry_strategy = Retry(
-        total=3,
-        status_forcelist=[429, 500, 502, 503, 504], 
-        allowed_methods=["HEAD", "GET", "OPTIONS"], 
-        backoff_factor=1
-    )
-
-    adapter = HTTPAdapter(max_retries=retry_strategy)
-    http = requests.Session()
-    http.mount("https://", adapter)
-    http.mount("http://", adapter)
-    return http
-
 class SoundCloudException(Exception):
     def __init__(self, code, msg):
         self.code = code
@@ -133,7 +116,7 @@ class SoundcloudAPI:
         return response.json()
 
     def update_playlist_details(self, playlist_id, title=None, description=None):
-        url = f"{self.base_url}/playlists/{playlist_id}"
+        url = f"{SOUNDCLOUD_URL}/playlists/{playlist_id}"
         data = {
             "playlist": {}
         }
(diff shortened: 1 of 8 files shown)


rimma-kubanova:
I updated this PR

* Integrated SoundCloud export (can successfully create playlists in SoundCloud and put tracks into it)

* SoundCloud lookup (still needs some tweaking, since lookup doesnt always finds songs. I guess it just needs more data to work with. OVERALL the logic is implemented)

amCap1712 (changes requested):
The api calls should use `linked_partitioning=true` where relevant, the format of the JSON response changes if you pass the query param so be sure to test the patches again and update the code that accesses the JSON. 

You can find an example here of how to paginate with linked_partitioning: https://github.com/metabrainz/listenbrainz-server/blob/f0334debe2b003cc36a6d94765442bf3be0a34d5/listenbrainz/metadata_cache/soundcloud/handler.py#L66-L109.

amCap1712:
Also, the code to use soundcloud export using CLI was missing. I added it myself  because I needed it for testing.

troi/tools/utils.py
```diff
@@ -18,3 +20,107 @@ def create_http_session():
...
+        """
+        retry_strategy = Retry(
+            total=3,
+            status_forcelist=[429, 500, 502, 503, 504], 
+            allowed_methods=["HEAD", "GET", "OPTIONS"], 
+            backoff_factor=1
+        )
+
+        adapter = HTTPAdapter(max_retries=retry_strategy)
+        http = requests.Session()
```
amCap1712, line 50: Add a hook that calls `raise_for_status` automatically after the API call.

rimma-kubanova:
@amCap1712 Thanks for the feedback! Working on it. Completely forgot about cli, will add them to others then

rimma-kubanova:
@amCap1712 hi! I updated the code

troi/tools/utils.py
```diff
@@ -18,3 +21,142 @@ def create_http_session():
...
+
+            if response.status_code != 200:
+                track_details.append({
+                    "id": track_id,
+                    "title": None,
+                    "playable": False
+                })
+                continue
+
+            data = response.json()
```
amCap1712, line 116: This looks wrong, track_details endpoint does not return a collection. You should create the track_data here directly from the response without any loop.

troi/tools/utils.py
```diff
@@ -18,3 +21,142 @@ def create_http_session():
...
+
+        return track_details
+
+    def get_playlist_tracks(self, playlist_id, **kwargs):
+        tracks = []
+        url = f"{SOUNDCLOUD_URL}/playlists/{playlist_id}/tracks"
+
+        while url:
+            response = self.session.get(url, headers=self.headers, params=kwargs)
+            response.raise_for_status()
```
amCap1712, line 139: raise_for_status call is redundant here because of the hook.

troi/tools/utils.py
```diff
@@ -18,3 +21,142 @@ def create_http_session():
...
+            data = response.json()
+            tracks.extend(data.get("collection", []))
+            url = data.get("next_href")
+
+        print(tracks)
+        return tracks
+
+    def get_unplayable_tracks(self, playlist_id, **kwargs):
```
amCap1712, line 147: This function is almost similar to the get playlist tracks one, you can call that function here to get the tracks in the playlist and then filter the unstreamable ones and return them.

troi/tools/utils.py
```diff
@@ -18,3 +21,142 @@ def create_http_session():
...
+        adapter = HTTPAdapter(max_retries=retry_strategy)
+        http = requests.Session()
+        http.mount("https://", adapter)
+        http.mount("http://", adapter)
+
+        def raise_for_status_hook(response, *args, **kwargs):
+            try:
+                response.raise_for_status()
+            except HTTPError as http_err:
+                print(f"HTTP error occurred: {http_err}")
```
amCap1712, line 59: Create a `logger = logging.getLogger(__name__)` at the top of the file and use `logger.error` here.

amCap1712 (changes requested):
Once these changes are done, I'll do a retest to ensure playlist exports work and then we can merge it.

rimma-kubanova:
@amCap1712 Thanks for your feedback! I've implemented all the changes you requested. 

I also made some additional tweaks and identified a few errors. The `fixup_soundcloud_playlist` function should now work properly, including handling unplayable tracks.

The only thing I'm facing is that the Labs Lookup api retrieves soundcloud ids for only a limited number of songs, which makes testing a bit difficult

PR #148, opened 2024-08-24
[bugfix]: Do not require Subsonic disc/track number, fix lb-radio artist locally

kgarner7:
This change has a few fixes:
- For subsonic import, allow track and disc number to be None (and default to 1). Subsonic servers are **not** guaranteed to return values here. Tested on my Navidrome instance
- Multiple fixes for get_similar_artists: proper post parameter, parse response, query using list of ids (not dicts), and return an empty array for msgs
- For `artist.py`, in some cases the `artist_credit` is `None`, also allow for this (via `AttributeError`)

These changes were tested locally by importing via subsonic, importing metadata, and then doing `lbz-radio` with an artist, and then with a tag.

### DIFF
--- a/troi/content_resolver/artist_search.py
+++ b/troi/content_resolver/artist_search.py
@@ -39,18 +39,15 @@ class LocalRecordingSearchByArtistService(RecordingSearchByArtistService):
 
         r = requests.post("https://labs.api.listenbrainz.org/similar-artists/json",
                           json=[{
-                              'artist_mbid':
-                              artist_mbid,
+                              'artist_mbids':
+                              [artist_mbid],
                               'algorithm':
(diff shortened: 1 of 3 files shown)


mayhem:
Thanks for that fix (and the unrelated fix that we should've caught, lol)! This is the first PR I've received on this code, so I am quite happy for that -- how are you finding these features??

kgarner7:
Someone in the Navidrome discord (https://discord.com/channels/671335427726114836/671335427726114839/1276785382553292892) found it and was having issues. I personally used troi before when someone in the team came in a while back and mentioned it, so I took a look, read the docs, and found a _different_ issue with Subsonic import (partially a Navidrome-specific detail/bug?).
Currently, while I can run it locally now, I still primarily use troi through the daily-jams playlists/listenbrainz API (rather than CLI).

mayhem:
Great, thanks for your response!

PR #150, opened 2024-09-02
SoundCloud import tracks fix

rimma-kubanova:
Thanks to @anshg1214 for highlighting this in the [PR](https://github.com/metabrainz/listenbrainz-server/pull/2965#pullrequestreview-2271886561). We noticed that importing playlists from SoundCloud wasn't functioning correctly

The issue stemmed from the endpoint output. After updating the API callback to support pagination, it returned tracks but omitted playlist details.

I discovered that the playlist endpoint doesn’t support pagination, so I implemented a separate endpoint specifically for fetching the playlist name and description.

This approach seems effective, and I’m considering applying a similar solution for other services


amCap1712 (commented):
LGTM, Thanks!

PR #151, opened 2024-09-05
LB-1628: LB radio 429 fix

mayhem:
LB-1628 points out that LB Radio is crap at dealing with rate limiting, so this PR ensures that request calls handle 439/503 correctly. 

Also, for some queries more than 1,000 items were generated, which caused problems. Truncating to first 1,000 tracks is a sufficient fix for this.

### DIFF
--- a/troi/content_resolver/artist_search.py
+++ b/troi/content_resolver/artist_search.py
@@ -2,6 +2,7 @@ import os
 from collections import defaultdict
 import datetime
 import sys
+from time import sleep
 
 import peewee
 import requests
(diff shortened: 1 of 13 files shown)


phw:
@mayhem Came here to check for issues because I faced the 1000 items issue for a lb-radio query. Tested the patch here and it is working well.

mayhem:
Thanks for this -- I already had a PR with these fixes, but then got distracted 17 ways.

mayhem:
Right, I'll take that as an approval, so I'll merge. :)

PR #152, opened 2024-12-03
Allow specifying subsonic playlist ID to overwrite

phw:
If given the playlist upload to subsonic will look for an existing playlist with the given ID and update it. If no playlist with the given ID is found a new playlist gets created.

This allows updating an existing playlist instead of creating a new one. E.g. I have troi generate me a new weekly jams playlist every Monday. I would like to have a single weekly jams playlist that updates every week, instead of having my playlists being spammed with a new separate entry every week. So I would run something like:

```
troi weekly-jams -ui 12c7a953-9ca8-48e4-9192-ec9c1e5a1761 outsidecontext
```

where `12c7a953-9ca8-48e4-9192-ec9c1e5a1761` is the subsonic ID of a playlist in my Navidrome instance.


mayhem (approved):
Very useful, thank you!

PR #153, opened 2024-12-04
Get all tracks from spotify playlist

MonkeyDo:
As reported in https://tickets.metabrainz.org/browse/LB-1514, the ImportPlaylistPatch is limited to 100 tracks per playlist, even if there are more in that playlist.
This is because the spotipy playlist_items method (and the underlying spotify api endpoint) only allows fetching 100 tracks at a time.
To get all the tracks in a playlist, we need to call this endpoint repeatedly with an offset updated each loop.


MonkeyDo:
Tested on test.LB and seems to have worked !
image

110 out of 169 songs, so something is not ideal, but certainly not limited to 100 tracks anymore

mayhem (approved):
Thanks!

PR #154, opened 2024-12-09
LB-1700: Get all tracks for apple music playlist

MonkeyDo:
Following #153, fixing the same issue for Apple Music.

The current implementation using the Apple Music API "?include=tracks" only returns the first 100 tracks with no pagination option.
To get the rest of the tracks if there are more, we need to fetch them separately using another endpoint ("/tracks"). Thankfully we can reuse the included tracks from the first call and start at offset 100.

### DIFF
--- a/troi/tools/utils.py
+++ b/troi/tools/utils.py
@@ -75,6 +75,15 @@ class AppleMusicAPI:
                 url = f"{APPLE_MUSIC_URL}/me/library/playlists/{playlist_id}/tracks?limit=100&offset={offset}"
                 response = self.session.get(url, headers=self.headers).json()
                 try:
+                    if "errors" in response:
+                        # https: // developer.apple.com/documentation/applemusicapi/errorsresponse
+                        error_objects = response["errors"]
+                        for error_object in error_objects:
+                            error_message = error_object["detail"] if "detail" in error_object else error_object["title"]
+                            logger.error(
+                                f"Error code {error_object['status']}: {error_message}")
+                        break
+
                     tracks_data = response["data"]
                     tracks.extend(tracks_data) if tracks_data else None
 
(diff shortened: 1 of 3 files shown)


troi/tools/utils.py
```diff
@@ -59,8 +59,36 @@ def playlist_add_tracks(self, playlist_id, track_ids):
...
+        playlist_description = response["data"][0]["attributes"].get(
+            "description", {}).get("standard", "")
+
+        # apple music returns only the first 100 tracks with "/playlists/{playlist_id}?include=tracks"
+        # and we need to fetch the rest of the tracks using the "playlists/{playlist_id}/tracks" endpoint
+        if len(tracks) < total_tracks_count:
+            offset = len(tracks)
+            # endpoint returns 100 tracks per call max -> run iteratively with an offset until there are no more tracks
+            while True:
+                url = f"{APPLE_MUSIC_URL}/me/library/playlists/{playlist_id}/tracks?limit=100&offset={offset}"
```
mayhem, line 75: Does any more error checking need to be added here? rate limiting or otherwise?

mayhem: We need to ensure that errors that might happen are not silently dropped.

mayhem (approved):
If there are no error handling improvements, good to merge.

troi/tools/utils.py
MonkeyDo: Rate limiting an retry strategy are handles by the session
https://github.com/metabrainz/troi-recommendation-playground/blob/6293c4993d40048492e20270269ec546b1b21be9/troi/tools/utils.py#L183C5-L197C16

What should be the strategy for handling other errors?
The API will sometimes return a json body with an `error` key, so I could try parsing for that and logging the error message (instead of the current caught KeyError) as an improvement.
Anything else that I should look out for?

mayhem: I think that is all good -- thanks!

PR #155, opened 2024-12-10
LB-1629, LB-1699: Truncate and clean up playlist description and title before exporting to Spotify

MonkeyDo:
Spotify limits the number of characters for a playlist title or description, and doesn't allow newlines.
Sending a description with newlines or too many characters results in a 500 error from the Spotify API.
HTML markup isn't respected either.

This PR removes HTML markup, newlines and truncates the strings to expected max lengths.

### DIFF
--- a/troi/tools/spotify_lookup.py
+++ b/troi/tools/spotify_lookup.py
@@ -1,6 +1,6 @@
 import logging
 from collections import defaultdict
-
+import re
 import requests
 import spotipy
 from more_itertools import chunked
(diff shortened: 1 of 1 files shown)


troi/tools/spotify_lookup.py
```diff
@@ -134,13 +134,24 @@ def submit_to_spotify(spotify, playlist, spotify_user_id: str, is_public: bool =
 
     logger.info("submit %d tracks" % len(spotify_track_ids))
 
+    # Truncate to character limit for title
+    playlist_name = playlist.name[0:100]
```
mayhem, line 138: This may look kinda weird. Maybe cut to 97 and then add "..." ? Fine to merge now, we can improve this later.

mayhem (approved):

troi/tools/spotify_lookup.py
MonkeyDo: Yes I'll opt for merging for now, I think 100+ chars is an edge case I'll be happy to fix if someone complains.

PR #156, opened 2024-12-11
Update user-guide.rst: resolve subsonic -> troi db subsonic

svanzoest:
`resolve subsonic` or `troi resolve subsonic` did not work for me, but looks like this is now `troi db subsonic`.

This still times out against my funkwhale server


mayhem (approved):
Thanks!

mayhem:
Perhaps you could turn up in our matrix channel so we can look at the timeout issue? https://musicbrainz.org/doc/Communication/ChatBrainz

PR #157, opened 2024-12-19
Update lb_radio.rst

zabe40:
The next section lists 7 different entities that are supported, not just artist and tag.


mayhem (approved):
Thanks!

PR #158, opened 2024-12-23
Fix subsonic upload_playlist not properly removing existing entries

phw:
The subsonic `upload_playlist` changes in #152 did not generate the removed indexes list correctly. Instead of the entries in the playlist it just counted the elements in the "playlist" structure, which is an object with various data about the playlist.

This change uses the `songCount` field to determine how many entries are in the playlist.

I tested this with subsonic, and now my weekly playlist is no longer growing every time it gets updated.


mayhem (approved):
Thanks!

PR #164, opened 2025-01-23
Hopefully fix CVE-2024-42471

mayhem:
Not entirely sure that is how you update version numbers here, we'll see.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: test
     steps:
-      - uses: actions/download-artifact@v4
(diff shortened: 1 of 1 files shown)


MonkeyDo (commented):
Latest version is [v4.1.8](https://github.com/actions/download-artifact/releases/tag/v4.1.8)
It might make sense to update to the latest while we're here, but i guess we don't have a set dependency upgrade policy.

mayhem:
Updated.

MonkeyDo:
I'm not sure this is entirely necessary, as the dependabot alert says both that the package is no longer vulnerable, and that the tag 'v4' should point to the latest version which is not vulnerable.

mayhem:
Well, lets do it anyway, since we're here. :)

PR #166, opened 2025-01-29
LB-1731: Create daily jams for users who don't have enough listens

mayhem:
As discussed in matrix....

### DIFF
--- a/troi/filters.py
+++ b/troi/filters.py
@@ -343,9 +343,12 @@ class LatestListenedAtFilterElement(troi.Element):
         :param min_number_of_days: The number of tracks that must have passed for a track to be kept.
     '''
 
-    def __init__(self, min_number_of_days=14):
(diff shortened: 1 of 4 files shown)


troi/filters.py
```diff
@@ -377,14 +386,17 @@ class NeverListenedFilterElement(troi.Element):
         Remove/keep only recordings if they have not/been listened to.
     '''
 
-    def __init__(self, remove_unlistened=True):
+    MIN_ITEMS_REQUIRED = 5
```
MonkeyDo, line 389: Just checking that this is expected to be defined again.
I see self.MIN_ITEMS_REQUIRED so I suppose it is a class const rather than a global one.

MonkeyDo (approved):
Looks good !
Thanks for the quick turnover!

PR #167, opened 2025-03-14
Minor cleanup

mayhem:
Fix a problem with unresolved releases, make quiet flag optional, and give loggers names, so that lb-local can capture the logs for display. Also, make the toolbar STFU duing quiet=True

### DIFF
--- a/troi/content_resolver/database.py
+++ b/troi/content_resolver/database.py
@@ -22,7 +22,7 @@ from troi.content_resolver.formats import mp3, m4a, flac, ogg_opus, ogg_vorbis, 
 
 from troi.content_resolver.utils import existing_dirs
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("troi_db_scan")
 
 SUPPORTED_FORMATS = (
     flac,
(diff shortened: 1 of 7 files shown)


troi/content_resolver/database.py
```diff
@@ -22,7 +22,7 @@
 
 from troi.content_resolver.utils import existing_dirs
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("troi_db_scan")
```
amCap1712, line 25: I think in general its better to retain the original name of the file, why do you want to modify the logger name here?

troi/content_resolver/database.py
```diff
@@ -353,7 +350,10 @@ def update_status(self, statusdata):
             Update status counter and display matching progress
         """
         self.counters.status[statusdata.status] += 1
-        if not self.quiet:
+
+        if self.quiet:
```
amCap1712, line 354: In future, I think it would be better to configure logging level based on the quiet flag in troi's entry points. Also, configure tqdm with it if possible. But should be fine for now.

amCap1712 (approved):
LGTM, Thanks!

troi/content_resolver/database.py
mayhem: so its a known and documented name, so that LB Local can capture the logger and its output for when LB local scans collections via troi.

PR #168, opened 2025-03-21
Improve http fetching

mayhem:
Using an HTTP session and a new rate limit http fetching module ensures that all requests are timed to not overload any of the APIs we're calling. Hack for MB is in place and will need to be removed once MB's RateLimit headers start working correctly. 

related: https://tickets.metabrainz.org/browse/MBH-589

### DIFF
--- a/tests/musicbrainz/test_mbid_mapping.py
+++ b/tests/musicbrainz/test_mbid_mapping.py
@@ -39,7 +39,7 @@ def mocked_requests_post(*args, **kwargs):
 
 class TestMBIDMapping(unittest.TestCase):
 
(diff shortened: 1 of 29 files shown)


troi/content_resolver/subsonic.py
```diff
@@ -54,8 +54,7 @@ def connect(self):
             self.config.SUBSONIC_HOST,
             username=self.config.SUBSONIC_USER,
             port=self.config.SUBSONIC_PORT,
-            salt=self.config.SUBSONIC_SALT,
```
mayhem, line 57: OOps, PR copypasta.

PR #182, opened 2026-02-21
LB-1919: Fix spotify lookups by using raw API calls to spotipy to handle recent API changes

meekhumor:
### Problem

`Spotify` recently introduced changes to several of its [web api endpoints](https://developer.spotify.com/documentation/web-api/references/changes/february-2026). As a result, the current version of the `spotipy` library is no longer fully compatible.

Specifically, some of [spotipy’s wrapper methods](https://github.com/spotipy-dev/spotipy/blob/master/spotipy/client.py) do not correctly parse these updated response structures, which causes playlist import and export functionality to fail.

### Temporary Solution

Until an updated version of `spotipy` is released, this issue can be mitigated by bypassing the library’s wrapper methods and making raw http requests to the `spotify web api`.

This approach ensures compatibility with the new spotify api changes, serving as a temporary workaround until official spotipy library support is available.

### DIFF
--- a/tests/test_playlist.py
+++ b/tests/test_playlist.py
@@ -85,35 +85,35 @@ class TestSpotifySubmission(unittest.TestCase):
             }
         ])
 
-        mock_requests.post("https://api.spotify.com/v1/users/test-user-id/playlists", json={
+        mock_requests.post("https://api.spotify.com/v1/me/playlists", json={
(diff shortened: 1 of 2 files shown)


amCap1712:
fixed by updating spotipy instead.

PR #184, opened 2026-03-10
feat: Add XSPF playlist import functionality

anshg1214:
In this PR, we introduce a new patch to import playlists in xspf format.

### DIFF
--- a/troi/patches/playlist_from_xspf.py
+++ b/troi/patches/playlist_from_xspf.py
@@ -36,9 +36,12 @@ class ImportXSPFPlaylistPatch(Patch):
 
         name, desc, _ = parse_xspf(xspf_content)
 
-        source = RecordingsFromXSPFElement(xspf_content=xspf_content)
+        lb_token = inputs.get("token")
+        source = RecordingsFromXSPFElement(
+            xspf_content=xspf_content, listenbrainz_token=lb_token
+        )
 
-        rec_lookup = RecordingLookupElement()
+        rec_lookup = RecordingLookupElement(auth_token=lb_token)
         rec_lookup.set_sources(source)
 
         pl_maker = PlaylistMakerElement(name, desc, patch_slug=self.slug())
--- a/troi/patches/playlist_from_xspf.py
+++ b/troi/patches/playlist_from_xspf.py
@@ -1,0 +1,47 @@
+from troi import Playlist
+from troi.patch import Patch
+from troi.playlist import RecordingsFromXSPFElement, PlaylistMakerElement
+from troi.musicbrainz.recording_lookup import RecordingLookupElement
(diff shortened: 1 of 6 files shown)


MonkeyDo:
Initial test of the LB-side PR that uses this patch, I imported [this XSPF playlist](https://gist.github.com/MonkeyDo/2a2f1f200d6e5157c74ec68e5af44ac0) successfully [into LB](https://listenbrainz.org/playlist/c16b14b6-b76a-45af-bad0-49912f3aa8b5/), albeit with a bunch of tracks missing (imported 38/50).

I wanted to report that some of the tracks that were not found are definitely in MB but the deburring/ simplification of the titles is probably not agressive enough for it to work well.
An example being "Running Up That Hill (A Deal With God) - 2018 Remaster" by Kate Bush, that we definitely have in MB. Not sure where the track resolution trips up but thought I'd mention it.

There is also a spotify URI attached to each track in this XPSF, so I wonder if we should also be supplementing with the spotify lookup patch where applicable...

I suppose we might want to eventually use the mapper v2 for this?

anshg1214:
Yes, once we move to the mapper v2 in LB, we can expect better performance.

MonkeyDo:
Bump from @amCap1712 , please review this one when you have some time, so we can merge its sister on the LB-server side :bow:

troi/patches/playlist_from_xspf.py
```diff
@@ -0,0 +1,50 @@
...
+        return "import-xspf-playlist"
+
+    @staticmethod
+    def description():
+        return "Import an XSPF playlist file into ListenBrainz"
+
+    def create(self, inputs):
+        xspf_content = inputs["xspf_content"]
+
+        name, desc, _ = parse_xspf(xspf_content)
```
amCap1712, line 37: You are passing twice, once here and once in Recording element which is wasteful. See how _deserialize_from_jspf does it in one pass.

troi/tools/xspf_lookup.py
```diff
@@ -0,0 +1,62 @@
...
+    name = root.findtext(f"{{{XSPF_NS}}}title") or "Untitled Playlist"
+    description = root.findtext(f"{{{XSPF_NS}}}annotation") or ""
+
+    tracks = []
+    track_list = root.find(f"{{{XSPF_NS}}}trackList")
+    if track_list is None:
+        return name, description, tracks
+
+    for track in track_list.findall(f"{{{XSPF_NS}}}track"):
+        recording_name = _fix_entities(track.findtext(f"{{{XSPF_NS}}}title") or "")
```
amCap1712, line 42: We should handle at least the fields that are documented for JSPF or handled in troi's JSPF implementation. Extra fields can probably go in additional metadata?

PR #186, opened 2026-06-18
Fix for updating existing recordings in local database

broman-pl:
Update for resolving conflicts on local database updates


MonkeyDo (commented):
Thanks for opening a PR!

I have one question/concern: when the Recordings change, should the RecordingMetadata and RecordiongTag rows be removed, as they contain metadata about the previous recording?
Or maybe updated with `process_recordings`?

PR #188, opened 2026-07-27
fix(subsonic): 

kaffeeringe:
resolve Nextcloud Music URL pathing, legacyAuth, and missing MBID lookup

Nextcloud's Music app doesn't use a port. So connections failed. Moreover their Subsonic API doesn't return MBIDs. I tried to fix both.

### DIFF
--- a/troi/content_resolver/subsonic.py
+++ b/troi/content_resolver/subsonic.py
@@ -1,6 +1,8 @@
 import datetime
 import logging
 import time
+import re
+import difflib
 import libsonic
 import musicbrainzngs
 
@@ -13,7 +15,6 @@ from troi.content_resolver.model.database import db
 from troi.content_resolver.model.recording import Recording, FileIdType
 from troi.content_resolver.utils import bcolors
 
-# MusicBrainz API Client initialisieren
 musicbrainzngs.set_useragent("TroiNextcloudResolver", "1.0", "https://github.com/metabrainz/troi")
 
 logger = logging.getLogger("troi_subsonic_scan")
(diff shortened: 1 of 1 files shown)


kaffeeringe:
You can now pull the database from Nextcloud Music and send Playlists to Nextcloud Music - without MBIDs in their API

kaffeeringe:
Now you don't have to resync the whole database, if there is a new album or the first run stops for some reason. It now asks for all albums and then only syncs, what isn't in the database.

troi/content_resolver/subsonic.py
```diff
@@ -9,7 +14,8 @@
 from troi.content_resolver.model.database import db
 from troi.content_resolver.model.recording import Recording, FileIdType
 from troi.content_resolver.utils import bcolors
-from troi.content_resolver.py_sonic_fix import FixedConnection
+
+musicbrainzngs.set_useragent("TroiNextcloudResolver", "1.0", "https://github.com/metabrainz/troi")
```
MonkeyDo: This seems very specific to nextcloud rather than a more generic Subsonic user agent

troi/content_resolver/subsonic.py
```diff
@@ -20,136 +26,214 @@ def applog(message, *args, **kwargs):
...
+
+    def _getApiUrl(self, action):
+        return f"https://{self._hostname}{self._serverPath}/{action}.view"
+
+    def _doInfoReq(self, req):
+        return super()._doInfoReq(req)
 
-    # Determined by the number of albums we can fetch in one go
+
+class SubsonicDatabase(Database):
```
MonkeyDo: We lost the docstring here:

```suggestion
class SubsonicDatabase(Database):
    '''
    Add subsonic sync capabilities to the Database
    '''
```

MonkeyDo:
Hello @kaffeeringe !
Thanks for looking into this.
Could I ask you to open separate PRs for each separate concern? The PR is too big to review properly and changes almost all of this file, making it hard to reason.
Separate PRs will allow us to review each functionality in isolation.

PR #189, opened 2026-07-30
feat(recording-lookup): add service hook to bypass HTTP call

anshg1214:
## Summary

Add a service hook to `RecordingLookupElement` so downstream consumers can bypass the HTTP call to `/1/metadata/recording` with a local implementation.

## Change

`RecordingLookupElement.read()` now checks for a registered `"recording-lookup"` service on the patch before making the HTTP POST. If found, calls `service.lookup(recording_mbids, inc)` instead. If not found, falls back to the existing HTTP path unchanged.

```python
lookup_service = self.patch.services.get("recording-lookup")
if lookup_service is not None:
    data = lookup_service.lookup(recording_mbids, inc)
else:
    # existing HTTP path, unchanged
```

## Backwards compatibility

Fully backwards compatible. No service registered = identical behaviour to today.

## Context

Companion PR in listenbrainz-server provides the `RecordingLookupService` implementation that calls `mapping.mb_metadata_cache` directly, eliminating an HTTP loopback to the same Flask server on every radio playlist generation.


MonkeyDo (approved):
LGTM 👍

PR #190, opened 2026-08-08
feat(lb-radio): add service hooks for stats, playlist and recs elements

anshg1214:
## Summary

Add service hooks to three LB Radio elements so downstream consumers can bypass HTTP API calls with local implementations. Companion to metabrainz/listenbrainz-server (see LB PR linked below).

## Change

Three elements updated to check for a registered service before making HTTP calls, following the same pattern as the existing `RecordingLookupElement` hook (#189):

**`LBRadioStatsRecordingElement`** -- checks for `"stats"` service before calling `liblistenbrainz.get_user_recordings()`.

**`LBRadioPlaylistRecordingElement`** -- checks for `"playlist"` service before calling `GET api.listenbrainz.org/1/playlist/{mbid}`. Stores `auth_token` directly on the element (previously only stored as `self.headers`).

**`LBRadioRecommendationRecordingElement`** -- checks for `"recs"` service before the paginated `liblistenbrainz.get_user_recommendation_recordings()` loop. Refactored to use `_select_recs()` helper so the filter and shuffle logic is shared between the service path and the HTTP fallback.

```python
service = self.patch.services.get("stats") if self.patch else None
if service is not None:
    raw_recordings = service.fetch(self.user_name, self.time_range, offset)
else:
    # existing HTTP path, unchanged
```

## Backwards compatibility

Fully backwards compatible. No service registered = identical behaviour to today.

## Context

Companion PR in listenbrainz-server provides the service implementations that call the DB directly, eliminating the remaining HTTP loopbacks in the LB Radio pipeline.


troi/patches/lb_radio_classes/playlist.py
```diff
@@ -29,31 +30,34 @@ def outputs(self):
...
-        r = http_get(f"https://api.listenbrainz.org/1/playlist/{self.mbid}", headers=self.headers)
-        if r.status_code == 404:
-            raise RuntimeError(f"Cannot find playlist {self.mbid}.")
-        if r.status_code != 200:
-            raise RuntimeError(f"Cannot fetch playlist {self.mbid}. {r.text}")
-
         # Give feedback about the playlist
         self.local_storage["data_cache"]["element-descriptions"].append(f"playlist {self.mbid}")
 
-        # Fetch the recordings, then shuffle
```
MonkeyDo, line 42: AI keeps removing comments that are acutally useful. Please keep the comments

troi/patches/lb_radio_classes/recs.py
MonkeyDo, line 1: Same here about the comments removal.

troi/patches/lb_radio_classes/recs.py
```diff
@@ -42,36 +56,35 @@ def read(self, entities):
...
+                for r in result['payload']['mbids']:
+                    if r.get("recording_mbid") is not None:
+                        offset += 1
+                        latest = r.get("latest_listened_at")
+                        if self.listened == "all" or \
+                                (self.listened == "unlistened" and latest is None) or \
+                                (self.listened == "listened" and latest is not None):
+                            count -= 1
+                            recordings.append(Recording(mbid=r["recording_mbid"]))
+
```
MonkeyDo: Shouldn't this use the new `_select_recs` method?

MonkeyDo (commented):
Small comments

MonkeyDo (approved):
👍

PR #191, opened 2026-09-03
fix: allow PUT method in http_fetch (LB-2016)

anshg1214:
# Problem

LB-2016

The SoundCloud data export uses a PUT request to update the export status, but `troi/http_request.py:http_fetch` only allows GET and POST methods. Any caller that passes `method="PUT"` gets a `ValueError`, which breaks the SoundCloud export flow.

# Solution

- Add `"PUT"` to the allowed methods whitelist
- Replace the `if/else` GET/POST dispatch with `session.request(method, ...)`, which handles any allowed method without needing a branch per verb

* [x] I have run the code and manually tested the changes

# AI usage

* [ ] I did not use any AI
* [x] I have used AI in this PR (add more details below)

#### If you did use AI:
* [ ] I used AI tools for communication
* [x] I used AI tools for coding
* [x] I understand all the changes made in this PR


MonkeyDo (approved):
Sorry, missed this one.
Looks good!

